PR #94, opened 2020-08-07
docs: improving timestamp concept illustrations

TheTanc:


docs/concept/designated-timestamp.md
```diff
@@ -55,11 +55,19 @@ existing timestamp columns. Note that:
...
+  src="/img/docs/concepts/designatedTimestamp.svg"
+/>
 
 Attempts to insert `out-of-order` timestamps will be rejected:
 
-![Diagram of an out of order insertion being rejected](/img/docs/concepts/timestampReject.jpg)
+<img
+  alt="Diagram of an out of order insertion being rejected"
+  className="screenshot--shadow screenshot--docs"
+  src="/img/docs/concepts/timestampReject.svg"
```
bluestreak01, line 69: I'm pretty sure camel case is not SEO friendly

TheTanc: Almost all images are camelCase currently. We can change all of them to kebab (should we?), but that will be in a separate PR.

mpsq: Name of image does not really matter, it just needs to be consistent. Tanc & I decided to go for camelCase, that's why

PR #96, opened 2020-08-10
docs: add WITH keyword

TheTanc:


docs/reference/configuration/server.md
```diff
@@ -15,7 +15,7 @@ configuration file as follows:
 ```shell
 aaa.bbb.ccc.ddd=myValue
 ```
-
+``
```
mpsq, line 18: is that intended?

sidebars.js
```diff
@@ -135,6 +135,7 @@ module.exports = {
             "reference/sql/timestamp",
             "reference/sql/truncate",
             "reference/sql/union",
+            "reference/sql/with",
             "reference/sql/where",
```
mpsq, line 139: with after where

docs/reference/sql/with.md
```diff
@@ -0,0 +1,40 @@
...
+sub-queries are used several times. 
+
+## Syntax
+
+ ![Flow chart showing the syntax of the WITH clause](/img/docs/diagrams/with.svg)
+ 
+ Where:
+ - `subQueryName` is the alias for the sub-query
+ - `subQuery` is a SQL query (e.g `SELECT * FROM table`)
+ - `mainQuery` is the main SQL query which involves the `subQuery` using its alias.
```
mpsq, line 19:
```suggestion
 - `mainQuery` is the main SQL query which involves the `subQuery` using its alias
```

docs/reference/configuration/server.md
TheTanc: nope

PR #103, opened 2020-08-17
chore(perf): do not split bundle in chunks

mpsq:


mpsq:
This was a test, now closing

PR #122, opened 2020-09-03
fix: CTA for blog post

davidgs:


blog/2020-08-19-questdb-performance.md
```diff
@@ -0,0 +1,258 @@
+---
```
mpsq, line 1: Please rebase your branch on master, I renamed the article (because of a non optimal URL), so the file should not be created be amended.

mpsq (changes requested):

davidgs:
Easier to fix in new PR.

PR #184, opened 2020-10-22
blog: update NYC taxi drivers article (correction by @nicquestdb)

mpsq:
Tweaks and fixes for the NYC taxis blog post, correction was 100% done by @NicQuestDB.

PR #250, opened 2020-12-02
docs(api): remove 400 error codes + fix curl queries - fix #184

bsmth:
__Description:__
In cases where API calls succeed but the query fails, we return a `200` with details in the response body for diagnosing the malformed request.

__Additional fixes:__
* multiple minor typos
* fixed CURL examples


docs/reference/api/rest.md
```diff
@@ -165,7 +166,6 @@ Considering the query:
 curl \
   -F schema='[{"name":"userId", "type": "STRING"},{"name":"movieId", "type":"STRING"}]' \
   -F data=@ratings.csv \
-  --data-urlencode "overwrite=true" \
```
bsmth, line 168: Side note: this fails for me using __curl 7.64.1 (x86_64-apple-darwin20.0)__:

```
Warning: You can only select one HTTP request method!
```

But this works fine:

```
curl -F data=@myfile.csv http://localhost:9000/imp\?overwrite\=true
```

docs/reference/api/rest.md
```diff
@@ -268,68 +267,34 @@ The HTTP status code will be set to `200` and the response will be:
...
 The parameters must be URL encoded.
 
-### Success response
+### Examples
 
 Considering the query:
 
 ```shell
-curl -G \
+curl -Gv \
```
mpsq, line 291: `v` is for `verbose` if I am correct, it is not a bad idea having it but we should be consistent, if we introduce `verbose` then we need to do it for all queries (or not, it is up to you, what you think makes more sense to users :) )

mpsq (changes requested):
:tada: congrats on your first PR!
It looks good, just a tiny remark regarding parameters passed to `curl`.

docs/reference/api/rest.md
```diff
@@ -268,68 +267,34 @@ The HTTP status code will be set to `200` and the response will be:
...
   --data-urlencode "query=select AccidentIndex2, Date, Time from 'Accidents0514.csv'" \
   --data-urlencode "limit=5" \
-  http://localhost:9000/exp
+  http://localhost:9000/exec
```
mpsq, line 294:
```suggestion
  http://localhost:9000/exp
```

This should be `exp`, this example is in the Export data section.

docs/reference/api/rest.md
```diff
@@ -355,37 +320,38 @@ The HTTP status code will be set to `200` and the response will be:
...
 
 Considering the query:
 
 ```shell
 curl -G \
-  --data-urlencode "query=select AccidentIndex2, Date, Time from 'Accidents0514.csv'" \
-  --data-urlencode "limit=5" \
-  http://localhost:9000/exp
+  --data-urlencode "query=SELECTT * FROM table;" \
+  http://localhost:9000/exec
```
mpsq, line 346:
```suggestion
  http://localhost:9000/exp
```

bsmth: Could be either here, right?

docs/reference/api/rest.md
bsmth: Right 👍

docs/reference/api/rest.md
bsmth: removed `-v` flag for simplicity

mpsq (approved):
LGTM - Nice work

PR #263, opened 2020-12-06
Update docker.md

prtkgpt:


mpsq:
The documentation does not support tags

PR #281, opened 2020-12-18
chore(copy): update quote

mpsq:
Wording/copy routine change


bsmth:
Looks like this is fixed in #282

bluestreak01:
already done

PR #293, opened 2020-12-28
Typo

TheTanc:


bluestreak01:
hey Tanc, the `&apos;` seems to be a valid HTML element. The already page renders as expected:

image

PR #304, opened 2021-01-11
feat: add tutorial section

mpsq:
This pull request adds a new "Tutorial" section. A tutorial can either be:

- A JSON object defined in `src/assets/tutorial.ts`, in the case the resource should be a link to a 3rd party website
- An internal page, hosted on questdb.io. In this case the page should be defined in `tutorial/` as an mdx file following the same conventions as a blog post

When a tutorial is `featured` (to set in the attributes at the top of the markdown page or as a property in the JSON object), it will appear in the top section, otherwise it will be in the "all" section. Both sections are sorted by date (defined in the file name for internal tutorials or in the JSON object otherwise).

This PR is still a draft and required checks/corrections/approval from @bsmth.


bsmth:
Looks good from my perspective, 👍

bluestreak01 (approved):

PR #308, opened 2021-01-12
docs: fix failing pgwire examples

bsmth:
__Description:__
Examples should not fail if re-run, this PR fixes postgres wire examples with the following changes:
* `create table` statements modified to include `if not exists` clause
* Prepared statements and parameterized queries added where possible


docs/develop/insert-data.md
```diff
@@ -533,39 +544,77 @@ pool.end().then(() => console.log("pool has ended"))
...
 }
 
-func checkErr(err error) {
+func preparedStatement() {
+	ctx := context.Background()
+	conn, err := pgx.Connect(ctx, "postgresql://admin:quest@localhost:8812/qdb")
+	defer conn.Close(ctx)
+
+	// Prepared statement has the name 'ps1'
+	_, err = conn.Prepare(ctx, "ps1", "insert into trades values ($1, $2)")
```
mpsq, line 578: Uppercase on SQL keywords for consistency

docs/develop/insert-data.md
```diff
@@ -391,12 +391,10 @@ func checkErr(err error) {
 
 </Tabs>
 
-## Postgres compatibility
+## PostgreSQL wire protocol
```
mpsq, line 394: The current wording was chosen as an attempt to be more transparent to users.
In reality, most Postgres users do not know what the "wire protocol" is and what are the implications. "Postgres compatibility" was supposed to convey the message clearly: "In a nutshell, we support Postgres clients". If you think it is not good, feel free to change it but keep in mind that you will have to update in other pages (in "develop") for consistency.

mpsq (changes requested):
Minor feedback, looks good overall

docs/develop/insert-data.md
bsmth: OK thanks for the clarification, I will revert to compatibility for the moment

bsmth:
@mpsq do you want to jump in to clean up the node (maybe Go also) examples? Happy to chuck the more complicated ones out to a different place if there's too much happening there

PR #318, opened 2021-01-18
docs(tutorial): add n8n

bsmth:
__Additions:__
* n8n bitcoin ticker tutorial

__Open Questions:__
-  [x] Nice way to add persistence when running questdb via docker


mpsq (approved):

tutorial/2021-01-18-low-code-bitcoin-ticker-workflow-with-time-series-database.mdx
```diff
@@ -80,7 +80,19 @@ This command mounts a volume with `-v` so that user configurations will be saved
 at `~/.n8n`. QuestDB can be started with the following docker command:
 
 ```
-docker run -p 9000:9000 -p 8812:8812 questdb/questdb
+docker run -p 9000:9000 -p 8812:8812 --name questdb questdb/questdb
+```
+
+By giving the container a name with `--name questdb`, we can reuse it at a later point and
```
mpsq, line 86: I am being picky but maybe it would be better if we do not use "questdb" as the container name? It could easily be confused with the name of the Docker image, or the name of the Docker org. Something like "mydb" or "questdb-tutorial", "n8n-tutorial, "questdb-n8n", etc. might be better.

mpsq (changes requested):
minor change

tutorial/2021-01-18-low-code-bitcoin-ticker-workflow-with-time-series-database.mdx
bsmth: good point, let me change 👍🏻

tutorial/2021-01-18-low-code-bitcoin-ticker-workflow-with-time-series-database.mdx
```diff
@@ -0,0 +1,226 @@
...
+
+This command mounts a volume with `-v` so that user configurations will be saved
+at `~/.n8n`. QuestDB can be started with the following docker command:
+
+```
+docker run -p 9000:9000 -p 8812:8812 --name questdb questdb/questdb
+```
+
+By giving the container a name with `--name questdb`, we can reuse it at a later point and
+the data that we have ingested will be persisted. If we want to re-run the same
```
mpsq, line 87: This sentence is not 100% correct:

1. Data is not persisted because of the use of the `--name` flag, Docker always persists changes in a container. `--name` is only a way to give the container a user friendly name the user can refer to rather than using Docker's auto generated aliases. I would just say: "By giving the container a name with `--name questdb`, we have an easy way to refer to the container created by `run` later on.".

2. Docker does not run images but containers, containers are created from an image, like an instance of an object. Rather than "If we want to re-run the same image after killing the running docker container, we can use the following commands:", I would say "If we want to re-use the same container (and its data!) after it has been stopped, we can use the following commands:".

mpsq (changes requested):
changes on Docker commands

tutorial/2021-01-18-low-code-bitcoin-ticker-workflow-with-time-series-database.mdx
bsmth: yep, great clarifications, thanks

mpsq (approved):

PR #334, opened 2021-01-26
chore: add youtube channel to navigation & footer links

bsmth:
__Background:__
We should add a link to the YouTube channel where we have been uploading webinars.


bsmth:
Q: Does it make more sense under Resources?

mpsq:
I would move the link to "resources", IMO, "community" is for content that has a social aspect, where people can interact with us.

bsmth:
Thanks, that was my gut feeling - I think it's more appropriate as  __Resources -> Videos__

mpsq (approved):
lgtm

PR #368, opened 2021-02-12
feat(index): add new integrations section + add link to enterprise

mpsq:


src/pages/index.tsx
```diff
@@ -722,7 +796,7 @@ const Home = () => {
       <Why />
       <Cards />
       <Console />
-      <SeenOn />
+      <Integration />
```
bsmth, line 799: I think the integrations could be higher up in the page

bsmth (dismissed):
additions look good 👍🏻

bluestreak01 (approved):

PR #369, opened 2021-02-12
docs: reword timestamp types in examples for consistency

bsmth:
__Description:__
* To reduce confusion with the use of **timestamp**, columns demonstrating timestamp type are named `ts`.
* Once-over for example uppercase/lowercase in SQL and illustrative tables

### DIFF
--- a/docs/reference/sql/where.md
+++ b/docs/reference/sql/where.md
@@ -222,7 +222,7 @@ designated timestamp can be applied
 SELECT scores WHERE ts = '2010-01-12T00:02:26.000Z';
 ```
 
-| timestamp                | score |
+| ts               | score |
 | ------------------------ | ----- |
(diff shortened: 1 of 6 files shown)


docs/reference/sql/where.md
```diff
@@ -222,7 +222,7 @@ designated timestamp can be applied
 SELECT scores WHERE ts = '2010-01-12T00:02:26.000Z';
 ```
 
-| timestamp                | score |
+| ts               | score |
```
mpsq, line 225: would it be possible to run prettier on that file?

mpsq (changes requested):

docs/reference/sql/where.md
bsmth: sure

mpsq (approved):
lgtm

PR #461, opened 2021-05-14
docs: 6.0.1 timestamp changes

ideoma:
__Changes:__

### Timestamp string equality
For `WHERE timestamp = '2020-01-01'` is not a range, but a single timestamp of `2020-01-01T00:00:00.000000Z`. The new equivalent with the semantics 'all data for this day' using string equality is

```sql
WHERE timestamp IN '2020-01-01'
```

### lg `<` and gt `>` operators

The string `2020-01-01` does not represent an interval, but a single `TIMESTAMP` data point of `2020-01-01T00:00:00.000000Z`

### `IN` list

To check if data within an interval, use `BETWEEN` i.e.: 

```sql
WHERE timestamp BETWEEN '2020-01-01T00:00:00.000000Z' AND '2020-01-02T00:00:00.000000Z' 
```

### DIFF
--- a/blog/2021-05-10-questdb-release-6-0-tsbs-benchmark.md
+++ b/blog/2021-05-10-questdb-release-6-0-tsbs-benchmark.md
@@ -1,5 +1,5 @@
 ---
-title: Building a new vector based storage model
+title: How we achieved write speeds of 1.4 million rows per second
 author: Vlad Ilyushchenko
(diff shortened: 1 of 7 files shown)


bsmth (approved):
Thanks @ideoma , I've made some minor formatting changes, updates 👍🏻

PR #479, opened 2021-06-09
chore: use theme's default secondary button style

bsmth:
Minor edit to the nav button to use the default `secondary` button style instead of hand-written values


bluestreak01:
can we use "BENCHMARK" all caps on the button? To make it consistent with all other buttons

bsmth:
> can we use "BENCHMARK" all caps on the button? To make it consistent with all other buttons

we can of course, this is overridden by the `text-transform` style

bluestreak01 (approved):

PR #485, opened 2021-06-13
Website Doc Fix: Invalid Docker volume mounting command

darthsiderius:
The websites getting started docs refers to a incorrect command to use mounted docker volumes with questdb. 

The mount point is "root/.questdb/db" not "root/.questdb/", which causes invalid behavior when using persisted volumes.


bsmth:
Thanks @darthsiderius for the PR. At the moment, this is intentional as a means of being able to pass in server config files to the container on startup, see https://github.com/questdb/questdb/blob/master/core/Dockerfile#L45-L55

bsmth:
Hi @darthsiderius as of 6.0.3, given the following directory structure:

```bash
tree -L 1
.
├── conf
├── db
└── public
```
it's suggested to mount as follows:

```
docker run -v "$(pwd):/root/.questdb/"  questdb/questdb
```

If you do not require passing in server config files (in `conf` dir) into the container, you may mount `db` only. Given the following directory contents:

```bash
tree -L 1
.
├── _tab_index.d
├── table_1
└── table_2
```

you may mount to `db` as follows:

```
docker run -v "$(pwd):/root/.questdb/db"  questdb/questdb
```

Thanks for raising the issue 🙌🏻 , but I will close this for now as we're documenting intended use. 

However, if you think this is a bug in how we are handling the docker image, feel free to report it in https://github.com/questdb/questdb/issues

darthsiderius:
Thanks for the reply and clarification . Love y'alls work. 

Understood. When using a server conf the mounting point is "$(pwd):/root/.questdb/".  If that is the case, I think we found a different bug. If so, we will report an issue. 

Best of luck.

PR #486, opened 2021-06-14
Missing app name for helm install

spxie85:


bsmth (approved):
Thanks @spxie85 for the PR!

PR #563, opened 2021-10-05
Fix cached config vars in Docusaurus

insmac:
**Issue:**
Changing config variables (`customFields`) in `docusaurus.config.js` is not reflected in the build, as they appear to be cached unless the actual application codebase gets changed too.

**Fix:**
Instead of using the Docusaurus Context hook to get the variables in React components, use plain JS files that can be used in React directly and in Docusaurus config as well.


bsmth (approved):
Looks good 👍🏻

bsmth (approved):
Thanks 👍🏻

PR #597, opened 2021-11-02
docs: document strpos(), left(), right() functions

puzpuzpuz:
Documentation counterpart for this core PR: https://github.com/questdb/questdb/pull/1506

Also includes the following:
* Document `left()`/`right()` functions
* Add a readme note on `autoconf` package for Linux users


docs/reference/function/text.md
```diff
@@ -74,6 +74,54 @@ SELECT name a, length(name) b FROM names limit 4
 | TOM    | 3   |
 | null   | -1  |
 
+## charindex
+
+`charindex(substring, string, start)` - searches for a substring in a string, and returns the position. If the substring is not found, this function returns `0`. The performed search is case-sensitive.
```
bsmth, line 79: Could you run `prettier` on this file, please?

docs/reference/function/text.md
```diff
@@ -74,6 +74,54 @@ SELECT name a, length(name) b FROM names limit 4
...
+
+**Arguments:**
+
+- `substring` is a string to search for.
+- `substring` is a string to search in.
```
bsmth, line 84: typo `string`

docs/reference/function/text.md
```diff
@@ -74,6 +74,54 @@ SELECT name a, length(name) b FROM names limit 4
...
+| name          | idx |
+| ------------- | --- |
+| Tim Thompson  | 5   |
+| Anna Thompson | 6   |
+| Anna Mason    | 0   |
+| Tom Johnson   | 0   |
+
+:::tip
+
+`charindex()` can be compined with `left()`/`right()` functions to split a string column into two separate columns. See an example below.
```
bsmth, line 106: The code example is good, but please remove the `:::tip` block and describe a bit more what's happening in the code example with an extra one or two sentences to guide readers a bit more, something like:

> Assuming we have a table `example_table` with a single string type column `col`:
> 
> | col          |
> | ------------ |
> | apple,pear   |
> | cat,dog      |
> | ...          |
> 
> We can use `charindex()` to split the string values of `col`, in this case splitting at the comma `,` character. By using `left()`/`right()` functions, we can choose the string values at the left and right of the comma:
> 
> ```questdb-sql title="Splitting string into two separate columns"
> SELECT col,
>        left(col, charindex(',', col) - 1) as col1,
>        right(col, length(col) - charindex(',', col)) as col2
> FROM (SELECT col FROM example_table);
> ```
> 
> | col          | col1  | col2 |
> | ------------ | ----- | ---- |
> | apple,pear   | apple | pear |
> | cat,dog      | cat   | dog  |

This way, we avoid introducing `rnd_str` and `long_sequence` into the mix and reduce complexity / cognitive load.

docs/reference/function/text.md
puzpuzpuz: Done in 2cd592ccaf6698c3ce0a470a429707d249aff437

docs/reference/function/text.md
puzpuzpuz: My bad with this leftover. Thanks for the heads-up. Fixed in 2cd592ccaf6698c3ce0a470a429707d249aff437

docs/reference/function/text.md
puzpuzpuz: I was following another `:::tip` example from this doc section, but probably it's an old way to document things. Updated according to your suggestion in 2cd592ccaf6698c3ce0a470a429707d249aff437

puzpuzpuz:
@bsmth I've included documentation for `left()`/`right()` functions in 4bb9c2ffdf596fd4184d231da4c9b99e01c6a36e. PTAL

docs/reference/function/text.md
bsmth: As a general guideline, we try to avoid overusing the admonition blocks, limiting to one or two per page. I would recommend using them for important internal links and for caution/warning cases, usually.

puzpuzpuz:
Renamed `charindex` (SQL Server-like function) into `strpos` (PostgreSQL-like function) in 214f83498a952318400de57e10b66e006aa2151a. The core PR will be updated accordingly.

puzpuzpuz:
@bsmth the core PR is now merged, so this one should be good to go. PTAL.

bsmth:
> @bsmth the core PR is now merged, so this one should be good to go. PTAL.

Thanks, having a look now 👀

bsmth (approved):
Thanks a lot, it looks good 👍🏻

puzpuzpuz:
Since v6.1.1 is released, this one should be good to go.

bsmth:
> Since v6.1.1 is released, this one should be good to go.

Yep, merging now, thanks a lot for the changes!

PR #607, opened 2021-11-12
chore: job position redirects and url changes

insmac:


netlify.toml
```diff
@@ -9,3 +9,11 @@
...
       "node_modules/.cache",
     ]
+
+[[redirects]]
+  from = "/careers/head-of-talent/"
+  to = "/careers/people-and-talent-acquisition/"
+
+[[redirects]]
+  from = "/careers/backend-software-engineer/"
+  to = "/careers/core-database-engineer/"
```
bsmth, line 19: We're managing redirects in https://github.com/questdb/questdb.io/blob/master/static/_redirects

Although this looks a lot cleaner and easier to read, it might be better to have all redirects in one place, what do you think?

bsmth (approved):
Thanks 👍🏻

PR #608, opened 2021-11-16
chore(docs): improve node.js pg snippets

puzpuzpuz:
Fixes timestamp/date types handling in PostgreSQL protocol insertion sample for Node.js. Namely, the following code
```js
      const query = {
        name: "insert-values",
        text: "INSERT INTO trades VALUES($1, $2, $3, $4);",
        values: [Date.now() * 1000, Date.now(), "node pg prep statement", rows],
      }
```
was leading to errors like the following one `error: could not parse [value='1637047205399000', as=TIMESTAMP, index=0]` on `pg` v8.7.1. That's because `pg` was sending the provided number as is, without converting it to a timestamp string in a proper format.

Also improves all `pg` snippets to make them more readable and consistent with the rest of snippets for node.


bsmth:
taking a look 👀

bsmth (approved):
Thanks 🙌🏻

PR #610, opened 2021-11-17
chore: move redirects to netlify toml

insmac:


bsmth (approved):
Tested a few URLs and works as expected 👍🏻

PR #612, opened 2021-11-19
chore(docs): clean up ILP prop names and their defaults

puzpuzpuz:
Removes/updates obsolete prop names in ILP docs and updates the defaults to actual values.


bsmth (approved):
Thanks a lot!

PR #616, opened 2021-11-23
chore(docs): improve munmap wording on storage page

puzpuzpuz:
Removes `(thus writing data to disk)` part of the text describing the append model. That's because on Linux `munmap` doesn't flush the pages to disk (note: it used to flush them on very old kernel versions). Alternatively, we could change it to `(thus **eventually** writing data to disk)`, but I'd rather remove this note completely as it might be a source of confusion.


bsmth (approved):
Thanks 👍🏻

PR #624, opened 2021-11-24
chore: add demo data sets

bsmth:
Example data sets were added to S3 for readers to be able to download and import example tables.


insmac (approved):
S3 URLs + all CURLs tested and working

PR #678, opened 2022-01-14
fix layout of SQL showcase on mobile screen

argshook:
Before:

2022-01-14-133758_screenshot

After:
2022-01-14-133808_screenshot


bsmth (approved):
Looks good 👍🏻

PR #681, opened 2022-01-18
implement local screenshot testing tool

argshook:
This PR implements a simple local screenshot testing tool.

PR includes a readme.md that describes how it works.

The code in this PR is separate from the rest of codebase, so it has no effect. I'm submitting this PR so that the code is not lost and can serve usefule purpose.

The purpose is to use this tool locally, on demand, whenever contributor (like me) wants to be sure if their changes are causing unexpected behiour.

In future it can be improved to also test on different viewports to ensure responsive design works as expected.

If this proves useful, eventually it would be a good idea to extract and reuse in other QuestDB pages.

Or, if it proves really very useful, it should be improved to be
included in CI pipeline, perhaps through a service like chromatic or applitools.

But for the time being, this relatively simple script can serve the purpose.


argshook:
for transparency: this PR is worth a repo, therefore it wont be merged but extracted. Closing

PR #682, opened 2022-01-18
fix `plugins/ts-compile.sh` after yarn upgrade

argshook:
updating dependencies brings in a different dependency tree which `plugins/ts-compile.sh` is not happy about.

this commit implements `tsconfig.json` for `plugins` with `skipLibCheck` to avoid the typescript type collision error

relevant to https://github.com/questdb/questdb.io/pull/681


bsmth (approved):
Thanks 👍🏻

PR #690, opened 2022-01-21
hide codeblocks for 500ms on initial load

argshook:
code blocks render with dark theme initially and once javascript kicks
in, their theme is changed to the one set by user.

In case user chooses light theme, code blocks appear with dark theme for
just a few frames and only during initial load.

However, that is enough to create an odd "flashing" effect when
opening some page with code blocks.

An ideal fix would be to serve user a page with appropriate theme in the
first place, i.e. if user chooses light theme, then all the assets are
tailored for light theme (as opposed to current implementation, which is
that all assets are available always).

Unfortunately, this is not a trivial change, it would require some server
side logic, and currently there is no server in play, it's all static files.

A non ideal workaround (this commit) is to hide code block elements for
500ms, give some time for their initial load, and then display.

Please note that this is only relevant for initial load. Once page is
loaded and user is navigating around by clicking buttons, none of this
is relevant, all themes work as expected.


bsmth:
The issue appears to persist on the preview & on localhost. FYI this behaviour is also present on the official project:

- visit https://docusaurus.io/docs/static-assets
- change the theme style
- refresh the page

argshook:
@bsmth thanks, Brian, yes the issue persist on preview, and it's because of "Content Security Policy", you should see such error in console:

2022-01-21-124936_screenshot

It's because i was naive and just added an inline script, whilst it
really should be within the bundle. I'll fix that and update this PR.

The issue is known by docusaurus, here's a comment from relevant source:

https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx#L39-L42

Nonetheless, i'd like to manage the workaround to work, cause that
flashing code block is not good experience IMO

insmac:
https://user-images.githubusercontent.com/1871646/150525542-10ad4f5e-71c3-45f6-929f-3c2cfd7f3280.mp4

The issue still persists on Chrome and PR preview.

argshook:
The last commit should work good enough.

By "good enough" i mean that it does it's job, but some flash can
sometimes still be visible, although it's much less frequent now.

insmac (approved):
It does the job, I don't think we can do more at this point without going much more low level.

PR #692, opened 2022-01-21
chore(ilp): ILP documentation update

### LINKED ISSUE #651: ILP documentation update
- emphasise limitation that ILP table names are case sensitive
- document duplicate column behaviour, which columns are selected in which scenarios
- document duplicate timestamp behaviour, which timestamp column is selected and when
- make it clear in the documentation that the field section can contain symbol types too (string without quotes will be parsed as symbol)

glasstiger:
Detailed description of the changes can be found in issue #651.

Closes #651


bsmth (approved):
Thanks a lot! 👍🏻

PR #693, opened 2022-01-21
feat: updated configuration reference doc to reflect refactoring of configuration properties

bluestreak01:


bsmth:
Thanks for the additions. One thing I noticed is that we have no indicators of the available options for params like setting `http.net.connection.sndbuf` to `2m`.

We could add a note somewhere that this can be `n<unit>`: where `<unit>` can be one of

* `m` - for MB
* `k` - for kB
* `?` - anything else?

And if there are any upper / lower limits for this. It would probably be enough to add this to the description column in the table for each param:

> Available units are `m` for MB, `k` - for kB.

bluestreak01:
thanks @bsmth I'll add a paragraph tomorrow and we would merge then

bsmth:
> add a paragraph tomorrow and we would merge

added in https://github.com/questdb/questdb.io/pull/693/commits/c3fc719c5e7bbadb0c82c3d64e46d45d937fb9a0

PR #695, opened 2022-01-24
prevent navbar right side to climb up on the left side

argshook:
This PR adds additional media query that shrinks the width of the search
bar in the navbar, to avoid situations like the following:

image


bsmth:
Still occurs on firefox (`95.0.2 (64-bit)`)
image

Alternatively, as the menu is getting cluttered, we could update the top nav to be slim and focused:

image

The convention with the top-10 other docusaurus sites seems to be

* Docs
* Blog
* Community
* API/Reference

__Examples:__
---
image
---
image
---
image
---
image

bsmth:
I still see this when page width between 1200 and 1700 on Firefox:

image

Another media query maybe?

bsmth (approved):
Thanks for the improvements! Let's look at the other display issues in a follow-up 👍🏻

PR #698, opened 2022-01-25
remove scrollbar css customization

argshook:
This PR removes scrollbar customization CSS in order to let the OS,
browser or user decide how scrollbar should look like.

On modern OS and browsers scrollbars are usually hidden after scrolling anyway.


bsmth (approved):
Looks good, tnx 👍🏻

PR #700, opened 2022-01-25
chore(ilp): ILP doc update

glasstiger:
Merging `line.tcp.default.partition.by` and `line.tcp.default.partition.by` into a new property.
`line.tcp.default.partition.by` will be used for both.


bsmth (approved):
Thanks 👍🏻

PR #709, opened 2022-02-02
update careers page

argshook:
This PR updates the careers page

- refactor repetitive HTML to instead loop over data structure
- update current open positions


src/pages/careers.tsx
```diff
@@ -92,108 +131,33 @@ const CareersPage = () => {
...
-                Details&nbsp;
-                <img
-                  alt="Right arrow"
-                  height={20}
-                  src="/img/pages/careers/arrowRight.svg"
-                  width={20}
-                />
-              </span>
-            </a>
+            {currentOpenings.map(({ href, title, location }) => (
```
bsmth, line 143: nice

bsmth (changes requested):
Maybe we should add redirects to the missing pages at

https://github.com/questdb/questdb.io/blob/master/netlify.toml

We can redirect to `/careers` maybe?

argshook:
> Maybe we should add redirects to the missing pages at
> 
> https://github.com/questdb/questdb.io/blob/master/netlify.toml
> 
> We can redirect to `/careers` maybe?

good idea, updated!

bsmth (approved):
Thanks 👍🏻

PR #711, opened 2022-02-03
remove `graceful-fs` & `makedir` devDeps

argshook:
This PR removes two dev dependencies that we can get away with:

`makedir` is not needed as replacement for `mkdir -p`, because node
natively supports `fs.mkdir(path, { recursive: true })`

`graceful-fs` is not needed, because the minimal usage of it does not
constitute a need for a dependency, when native node can handle our needs


bsmth (approved):
Thank you!

PR #712, opened 2022-02-03
chore: removed rnd_long256 arguments section

bziobrowski:
rnd_long256() doesn't accept 3 arguments mentioned in documentation.
I found a rnd_long256(int) version in source code but it seems meant for testing .


bsmth (approved):
Thanks, Bolek 👍🏻

PR #717, opened 2022-02-09
update developers count

insmac:
Update count in About Us page. 

TODO: Automate using Slack API, if it's worth it, considering site needs redeployment to refetch and it's a small task to manually adjust the number.


bsmth (approved):
Tnx 👍🏻

PR #719, opened 2022-02-09
Send GET requests to /exec endpoint.

nslocum:
While both GET and POST requests work, the docs state that `/exec` expects a GET request.
All the non-Python code examples already send GET requests.


bsmth (approved):
Thanks Nick 👍🏻

PR #720, opened 2022-02-10
chore(docs): add troubleshooting guide

puzpuzpuz:
Adds new troubleshooting page.


bsmth:
I think we should also add:

* https://stackoverflow.com/questions/65976247/how-to-delete-a-few-rows-via-sql-in-questdb/65985165#65985165
* https://stackoverflow.com/questions/67785629/what-does-max-txn-txn-inflight-limit-reached-in-questdb-and-how-to-i-avoid-it
* https://stackoverflow.com/questions/68194952/how-do-you-get-prometheus-metrics-from-questdb

puzpuzpuz:
@bsmth 

https://stackoverflow.com/questions/67785629/what-does-max-txn-txn-inflight-limit-reached-in-questdb-and-how-to-i-avoid-it is a bug, so I don't think we should include it. As for https://stackoverflow.com/questions/68194952/how-do-you-get-prometheus-metrics-from-questdb, we have a nice guide for it and all users should be able to find it easily. Added the two remaining items in e60cefd9c0cf9047912a68e1bbede0791719ccd6.

Thanks for the suggestions!

bsmth (approved):
Great start!

PR #724, opened 2022-02-10
docs: The default value for http.json.query.float.scale is 4.

nslocum:
The default value is 4:
https://github.com/questdb/questdb/blob/c7f0a9da5505264e4b88b9046e5762a224e8a8ce/pkg/ami/marketplace/assets/server.conf#L89


puzpuzpuz (approved):

puzpuzpuz:
Thanks for the fix!

PR #730, opened 2022-02-17
chore(docs): improve LATEST BY page

puzpuzpuz:
Includes the following changes:
* Renames LATEST BY to LATEST ON everywhere in the docs. Keeps a redirect for the old `/latest-by` page address.
* Adds a clarifying text for the LATEST ON intro text.
* Adds a new section for LATEST ON over sub-query.


bluestreak01:
I feel the opening paragraph could better explain why we can't just use `LIMIT -1`.

>LATEST ON PARTITION BY is used as part of a [SELECT statement](https://deploy-preview-730--unruffled-blackwell-31bfb2.netlify.app/docs/reference/sql/select/) for returning the most recent records per unique column value, commonly on STRING and SYMBOL column types. For the sake of brevity we to refer to this clause as LATEST ON.

I'd like to suggest something along the lines of:

```
For scenarios where multiple time series are stored in the same table it is relatively difficult to identify last items of these time series with standard SQL syntax. We introduces `LATEST ON` extension for `SELECT` clause to remove boilerplate clutter and splice table with relative ease. 
```

WDYT?

bluestreak01:
Directly after syntax railroad diagram we should attempt to document syntax components. In similar way method args are documents along side what method does.

bluestreak01:
Here:

image


Can we say "The old `LASTEST BY` syntax..." ?

bluestreak01:
The document lacks description of how to time-slice, the last paragraph `LATEST ON first` does show mix of `where` clauses, but it is more powerful with time-slicing.

bluestreak01:
We also should list all types, which can be used in `PARTITION BY` clause. I would also like to see description of how `LASTEST ON` will behave with non-symbol types and some advice around using timestamp to limit the performance impact

puzpuzpuz:
> The document lacks description of how to time-slice, the last paragraph `LATEST ON first` does show mix of `where` clauses, but it is more powerful with time-slicing.

Addressed all comments except for this one in 73439d10637bfca7f875fa8300be66c15a03b651

puzpuzpuz:
Addressed the remaining comment in 67ef3b743bd3da6ef3ca27151c6207a7cdc16ad3

docs/reference/sql/latest-on.md
```diff
@@ -36,6 +51,16 @@ LATEST ON pickup_datetime PARTITION BY payment_type;
 | Cash         | 2019-06-30T23:59:54.000000Z | 2             |
 | Card         | 2019-06-30T23:59:56.000000Z | 1             |
 
+The above query returns the latest value within each time series stored in the
+table. Those time series are determined based on the column(s) specified in the
```
bluestreak01, line 55:
```suggestion
table. Those time series are determined based on the values in the column(s) specified in the
```

bluestreak01 (changes requested):

docs/reference/sql/latest-on.md
```diff
@@ -1,21 +1,36 @@
...
+  Reference documentation for using LATEST ON keywords with examples for
   illustration.
 ---
 
-`LATEST ON PARTITION BY` is used as part of a
-[SELECT statement](/docs/reference/sql/select/) for returning the most recent
-records per unique column value, commonly on `STRING` and `SYMBOL` column types.
-For the sake of brevity we to refer to this clause as `LATEST BY`.
+For scenarios where multiple time series are stored in the same table it is
+relatively difficult to identify the lastest items of these time series with
```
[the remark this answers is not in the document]
puzpuzpuz: No need for clarification, it's a typo. Thanks for spotting it. Fixed in 411eb87f12908c79fbcefdc59e3db3b97348f829

docs/reference/sql/latest-on.md
puzpuzpuz: Applied in 411eb87f12908c79fbcefdc59e3db3b97348f829

bluestreak01 (approved):

PR #737, opened 2022-02-28
doc: added reference on mixed timestamp

bluestreak01:
Added reference to columnset description of timestamp, which has examples and explanation of how mixing designated timestamp works.


docs/reference/api/ilp/overview.md
```diff
@@ -209,3 +209,9 @@ tracking,loc=north val=200i 1000000000\n
 ```shell title="Example of ILP message sans timestamp"
 tracking,loc=north val=200i\n
```
glasstiger, line 210: although integer also works we could change this to val=200t, just to avoid confusion.

glasstiger (approved):

docs/reference/api/ilp/overview.md
bluestreak01: this is an example without explicit timestamp (sans timestamp), `val` is of type `long`

PR #744, opened 2022-03-01
add `theme-color` meta tag with dracula background value

argshook:


argshook:
This PR was an attempt at fixing browser header on iOS devices. In some instances browser header on mobile was rendered in pink color, same as the background color `star us on github` banner.

This color is not intentional, browser takes it automatically. There's a way to set color explicitly, which was tried in this PR, but it didn't work well enough.

Therefore, closing the PR until we revisit this issue again.

PR #747, opened 2022-03-07
chore: document types nullability

puzpuzpuz:
Also adds a note for lack of arithmetic operations support on LONG256 type.


docs/reference/sql/datatypes.md
```diff
@@ -6,23 +6,33 @@ description: Data types reference documentation.
...
+| `binary`          | `64+n*8`     | Yes      | Length-prefixed sequence of bytes whose length is stored as signed 64-bit integer with maximum value of `0x7fffffffffffffffL`.                                                                                                                                      |
+| `long256`         | `256`        | Yes      | Unsigned 256-bit integer. Does not support arbitrary arithmetic operations, but only equality checks. Suitable for storing hash code, such as crypto public addresses.                                                                                              |
+| `geohash(<size>)` | `8`-`64`     | Yes      | Geohash with precision specified as a number followed by `b` for bits, `c` for chars. See [the geohashes documentation](/docs/concept/geohashes/) for details on use and storage.                                                                                   |
+
+:::info
```
bsmth, line 27: I think it might be worth adding a new section instead of this admonition, something like this:

```
## Type nullability

Numeric nullable types, such as `int` or `double`, use a special number to mark
`NULL` values. Signed integer types use the minimal number from the allowed
range to mark a `NULL`. Here are the values used for `NULL` by type:

* __`int`:__ `0x80000000`
* __`long`:__ `0x8000000000000000L`
* __`double`:__ `NaN`

Because floating point types use the `NaN` value, `NULL` values of these
types still occupy disk space.
```

docs/reference/sql/datatypes.md
```diff
@@ -6,23 +6,33 @@ description: Data types reference documentation.
...
+
+Numeric nullable types, such as `int` or `double`, use a special number to mark
+`NULL` values. Signed integer types use the minimal number from the allowed
+range to mark a `NULL`, e.g. `0x80000000` for `int` or `0x8000000000000000L` for
+`long`. Floating point types use the `NaN` value. Thus, `NULL` values of these
+types still occupy disk space.
+
+:::
 
 :::info
```
bsmth, line 37: we should consider merging multiple admonitions of the same type, or make a new section `## Field sizes`

bsmth (approved):
Thanks for the additions 👍🏻 

See my comments before merging

docs/reference/sql/datatypes.md
puzpuzpuz: Good idea. Introduced a new section in ecd814531af57c38b965cc81b4f5b8e9bdce94dd

PR #754, opened 2022-03-11
doc: async offload configuration

bluestreak01:
Configuration properties for async offload


puzpuzpuz:
@bluestreak01 I've pushed an update for the new prop introduced by https://github.com/questdb/questdb/pull/1950. Also updated the descriptions to make them slightly more consistent.

puzpuzpuz (approved):

puzpuzpuz:
I've updated the PR according to the latest changes.

PR #770, opened 2022-03-21
feat: add OS error codes page

puzpuzpuz:
Also changes the docs structure: FAQ -> Troubleshooting is now Troubleshooting -> FAQ. That's because this section now also contains the error code list.


puzpuzpuz:
@bluestreak01 this one seems to be good to go. Could you merge it?

PR #775, opened 2022-03-25
Request to update typo

pradeephubgit:
In the 'https://questdb.io/docs/reference/command-line-options/' page, in the first table, in the description of Option '-t',  it we can replace 'If this option omitted' with 'if this option is omitted'. If you agree, please feel free to update.

- Current: If this option omitted.
- Suggestion: If this option is omitted. (I added 'is' to the sentence).

Thanks,
Pradeep.


puzpuzpuz (approved):
Thanks for spotting the misprint and fixing it!

PR #788, opened 2022-03-29
Request to update command for viewing table

pradeephubgit:
Thanks for the very nicely written document. 
After creating a table and inserting data into it, I added the 'sensors'; command for viewing the table. New users might benefit from it. If you agree, pls accept this change.

**Note** - In the 'Files changed' tab, two files are mentioned. One of them, 'blog/2022-01-27-release-sql-jit-compiler.md' was updated for another PR, but appears here as well. I am noting it here, so as to avoid confusion.

### DIFF
--- a/docs/get-started/first-database.md
+++ b/docs/get-started/first-database.md
@@ -66,14 +66,21 @@ about the functions used here, see the
 [row generator](/docs/reference/function/row-generator/) pages.
 
 Our `sensors` table now contains 10,000 randomly-generated sensor values of
-different makes and in various cities. It should look like the table below:
+different makes and in various cities. Use this command to view the table:
+
+
+```questdb-sql
+'sensors';
+```
+
+It should look like the table below:
 
 | ID  | make              | city     |
 | --- | ----------------- | -------- |
-| 1   | RS Pro            | New York |
-| 2   | Honeywell         | Chicago  |
-| 3   | United Automation | Miami    |
-| 4   | Honeywell         | Chicago  |
+| 1   | Honeywell         | Chicago  |
+| 2   | United Automation | Miami    |
+| 3   | Honeywell         | Chicago  | 
+| 4   | Omron             | Miami    | 
 | ... | ...               | ...      |
 
 Let's now create some sensor readings. In this case, we will create the table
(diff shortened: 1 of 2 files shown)


docs/get-started/first-database.md
```diff
@@ -66,14 +66,21 @@ about the functions used here, see the
...
+
+It should look like the table below:
 
 | ID  | make              | city     |
 | --- | ----------------- | -------- |
-| 1   | RS Pro            | New York |
-| 2   | Honeywell         | Chicago  |
-| 3   | United Automation | Miami    |
-| 4   | Honeywell         | Chicago  |
+| 1   | Honeywell         | Chicago  |
```
pswu11, line 80: This is fine but not necessary, be aware that the original section is not wrong - the table will look different each time since it's randomly generated.

pswu11 (approved):

docs/get-started/first-database.md
pradeephubgit: Oh, okay. Thanks for taking the time to review and providing guidance. I will certainly remember this.

docs/get-started/first-database.md
```diff
@@ -66,14 +66,21 @@ about the functions used here, see the
...
 Our `sensors` table now contains 10,000 randomly-generated sensor values of
-different makes and in various cities. It should look like the table below:
+different makes and in various cities. Use this command to view the table:
+
+
+```questdb-sql
+'sensors';
+```
+
+It should look like the table below:
```
argshook, line 76: as @pswu11 is suggesting, the table will not look the same for every user. Perhaps let's add a note exactly about it, what do you think?

```suggestion
Data is generated randomly, but the table should look similar to the one below:
```

argshook: Ah, the PR got merged, I was too slow with this comment :)

That's okay, if anyone would think my suggestion should get in, feel
free to open another PR.

pradeephubgit: Thanks so much, Arijus @argshook. Your enhancement makes it very clear to users. I will update in a new PR since this has got merged. Is that fine with you?

argshook: Hey, @pradeephubgit absolutely, feel free to open a PR, even if it's a small change.

PR #791, opened 2022-03-30
chore: update [why questdb]

pswu11:
- reduced the text length cuz long text breaks the style (quite ugly)
- CTA button is universal, now it's linked to github repo (install questdb)


src/pages/index.tsx
```diff
@@ -105,39 +105,38 @@ const FeatureTabs = () => {
...
-              <p className={prCss.property}>Postgres compatible</p>
+              <p className={prCss.property}>Apache License 2.0</p>
+              <p className={prCss.property}>Thriving developer community</p>
+              <p className={prCss.property}>Transparent development</p>
+              <p className={prCss.property}>Embedded in Java applications</p>
             </div>
 
-            <Button className={meCss.menu__cta} to="/enterprise">
-              Enterprise &gt;
+            <Button className={meCss.menu__cta} to="https://github.com/questdb/questdb#try-questdb">
```
argshook, line 138:
```suggestion
            <Button className={meCss.menu__cta} to="https://questdb.io/get-questdb/">
```

Should `Get Started` point to github? We have a similar page available. Wdyt?

pswu11: @argshook the reasoning is that - for gaining github stars, we need to send traffic to github repo. We're directing people here and there but after all, maybe it's easier if we focus on one place?

argshook (approved):

PR #792, opened 2022-03-30
Add a table containing the NULL values for nullable types

marregui:
and as an afterthought mention `IS NULL` and `IS NOT NULL`.


docs/reference/sql/datatypes.md
```diff
@@ -34,14 +34,41 @@ characters) or disk size, whichever is smaller.
 
 ## Type nullability
 
-Numeric nullable types, such as `int` or `double`, use a special number to mark
-`NULL` values. Signed integer types use the minimal number from the allowed
-range to mark a `NULL`. Here are the values used for `NULL` in numeric types:
+Nullable types use a specific value to mark `NULL` values.
```
pswu11, line 38: Maybe add a sentence like `Here are ...` or `The table below shows...` ?

docs/reference/sql/datatypes.md
```diff
@@ -34,14 +34,41 @@ characters) or disk size, whichever is smaller.
...
-range to mark a `NULL`. Here are the values used for `NULL` in numeric types:
+Nullable types use a specific value to mark `NULL` values. 
 
-- **`int`:** `0x80000000`
-- **`long`:** `0x8000000000000000L`
-- **`float`:** `NaN`
-- **`double`:** `NaN`
+| Type Name          | Null value             | Description                                                                                                           |
+|--------------------|------------------------|-----------------------------------------------------------------------------------------------------------------------|
+| `float`            | `java.lang.Float.NaN`  | Equivalent to `Float.intBitsToFloat(0x7fc00000)`.                                                                     |
```
puzpuzpuz, line 41: We should avoid any Java-only references in the docs. It's better to refer to IEEE 754 rather than insert Java code like `Float.intBitsToFloat(0x7fc00000)`. The same suggestion applies to the below items.

docs/reference/sql/datatypes.md
```diff
@@ -34,14 +34,41 @@ characters) or disk size, whichever is smaller.
...
+
+Alternatively, use the equality operator aliases: 
+
+```questdb-sql
+SELECT * FROM <table> WHERE <column> IS NULL;
+SELECT * FROM <table> WHERE <column> IS NOT NULL;
+```
+
+:::note
+An **important consequence** of this is that `NULL` values
```
puzpuzpuz, line 72: nit: `An **important consequence** of this` could be removed since it has no context now.

docs/reference/sql/datatypes.md
```diff
@@ -34,14 +34,41 @@ characters) or disk size, whichever is smaller.
...
 
-- **`int`:** `0x80000000`
-- **`long`:** `0x8000000000000000L`
-- **`float`:** `NaN`
-- **`double`:** `NaN`
+| Type Name          | Null value             | Description                                                                                                           |
+|--------------------|------------------------|-----------------------------------------------------------------------------------------------------------------------|
+| `float`            | `java.lang.Float.NaN`  | Equivalent to `Float.intBitsToFloat(0x7fc00000)`.                                                                     |
+| `double`           | `java.lang.Double.NaN` | Equivalent to `Double.longBitsToDouble(0x7ff8000000000000L)`.                                                         |
+| `long`             | `0x8000000000000000L`  | Min possible value a long can take -2^63.                                                                             |
```
puzpuzpuz, line 43:
```suggestion
| `long`             | `0x8000000000000000L`  | Minimal possible value a `long` can take, -2^63.                                                                             |
```

docs/reference/sql/datatypes.md
```diff
@@ -34,14 +34,41 @@ characters) or disk size, whichever is smaller.
...
+| `long256`          | `0x8000000000000000L`  | **x4**, *four consecutive long null values*.                                                                          |
+| `date`             | `0x8000000000000000L`  |                                                                                                                       |
+| `timestamp`        | `0x8000000000000000L`  |                                                                                                                       |
+| `int`              | `0x80000000`           | Min possible value an int can take, -2^31.                                                                            |
```
puzpuzpuz, line 47:
```suggestion
| `int`              | `0x80000000`           | Minimal possible value an `int` can take, -2^31.                                                                            |
```

docs/reference/sql/datatypes.md
```diff
@@ -34,14 +34,41 @@ characters) or disk size, whichever is smaller.
...
+| Type Name          | Null value             | Description                                                                                                           |
+|--------------------|------------------------|-----------------------------------------------------------------------------------------------------------------------|
+| `float`            | `java.lang.Float.NaN`  | Equivalent to `Float.intBitsToFloat(0x7fc00000)`.                                                                     |
+| `double`           | `java.lang.Double.NaN` | Equivalent to `Double.longBitsToDouble(0x7ff8000000000000L)`.                                                         |
+| `long`             | `0x8000000000000000L`  | Min possible value a long can take -2^63.                                                                             |
+| `long256`          | `0x8000000000000000L`  | **x4**, *four consecutive long null values*.                                                                          |
+| `date`             | `0x8000000000000000L`  |                                                                                                                       |
```
puzpuzpuz, line 45:
```suggestion
| `date`             | `0x8000000000000000L`  |  Minimal possible value a `long` can take, -2^63.                                                                                                                     |
```

docs/reference/sql/datatypes.md
```diff
@@ -34,14 +34,41 @@ characters) or disk size, whichever is smaller.
...
+| `timestamp`        | `0x8000000000000000L`  |                                                                                                                       |
```
puzpuzpuz, line 46:
```suggestion
| `timestamp`        | `0x8000000000000000L`  |    Minimal possible value a `long` can take, -2^63.                                                                                                                   |
```

docs/reference/sql/datatypes.md
```diff
@@ -34,14 +34,41 @@ characters) or disk size, whichever is smaller.
...
+| `int`              | `0x80000000`           | Min possible value an int can take, -2^31.                                                                            |
+| `symbol`           | `0x80000000`           | Symbols are stored as `int` offsets in a lookup file.                                                                 |
```
puzpuzpuz, line 48: nit: this document should describe user-facing behavior, i.e. help users to avoid possible collisions with `NULL`s when inserting their data. As far as I know, it's not possible for the user to access the symbol key (`int` code), so I'm not sure if they need this row.

The same consideration applies to STRING and BINARY (and maybe other?) types.

docs/reference/sql/datatypes.md
marregui: the document's opening line **The type system is derived from Java types.** so it only begets showing the exact value. I dispute this comment :)

docs/reference/sql/datatypes.md
marregui: thank you, amended... would `minimum` be a better alternative?

```
/**
 * The {@code Long} class wraps a value of the primitive type {@code
 * long} in an object. An object of type {@code Long} contains a
 * single field whose type is {@code long}.
...
 * @since   1.0
 */
@jdk.internal.ValueBased
public final class Long extends Number
        implements Comparable<Long>, Constable, ConstantDesc {
    /**
     * A constant holding the minimum value a {@code long} can
     * have, -2<sup>63</sup>.
     */
    @Native public static final long MIN_VALUE = 0x8000000000000000L;
```

docs/reference/sql/datatypes.md
marregui: well, I hope our documentation adds a whole lot more of technical detail in general. I cannot document the stored null value for only some nullable types and leave the reader wondering. I find it useful to, at least somewhere, state what is happening under the scenes regarding all types when it comes to null. If this helps me, it will help others. I dispute this comment.

puzpuzpuz: Documenting internals is certainly useful as long as we're certain that they never change or we clearly highlight that those are internal details, but this particular documentation section was supposed to describe NULL handling in the public API.

As for disputing this comment, no need to do that. It's a nit, so it's up to you to address it or not.

docs/reference/sql/datatypes.md
puzpuzpuz: Minimum is fine, of course.

docs/reference/sql/datatypes.md
puzpuzpuz: Java type system is partly derived from standards supported by HW and other programming languages. So, it's much better to refer to IEEE 754, rather than leave the user with a Java code snippet or a reference to a standard Java class. QuestDB is a database, so it should document its API and avoid leaking too much of the implementation language.

As for disputing the comment, I leave it up to you.

puzpuzpuz (approved):
LGTM in general. See the disputed comments for potential improvements.

docs/reference/sql/datatypes.md
marregui: :) I have amended so that it is a win win

marregui:
Thank you @puzpuzpuz for the review!, I have amended NaN, while still leaving a parens explaining where it comes from

pswu11:
would it be helpful if we point out that this is available since 6.3? some users might still be using the older version and intended to do the same according to docs

puzpuzpuz (approved):

PR #825, opened 2022-04-21
fix: status code changed to indicate errors

HeZean:
When an error occurs on a request via the REST API, the status code is set to `400`.

See questdb/questdb#2039.


bluestreak01 (approved):

HeZean:
Hi Vlad,

Sorry but I don't understand, why could changing one line in the document cause the build failed? Really appreciate if you can point it out.

PR #846, opened 2022-05-16
chore: added example of formatting year in multiple ways

bluestreak01:


docs/reference/function/date-time.md
```diff
@@ -966,32 +966,59 @@ Format is a combination of letters from table below combined with arbitrary
...
+| `yyy-MM`  | `1995-03-01T00:00:00.000000` | `995-03`     |
+| `yyyy-MM` | `1995-03-01T00:00:00.000000` | `1995-03`    |
+
+Let's try to parse text. 
+
+| Pattern   | Text      | Parse result                 | Remarks                                      |
+|-----------|-----------|------------------------------|----------------------------------------------|
+| `y-MM`    | `1995-03` | `1995-03-01T00:00:00.000000` | year is parsed greedily until next separator |
+| `y-MM`    | `995-03`  | `0995-03-01T00:00:00.000000` | year is parsed greedily until next separator |
+| `y-MM`    | `95-03`   | `2095-03-01T00:00:00.000000` | acts as `yy`                                 |
```
bziobrowski, line 1016: I what sense does  'y-MM' in 
```
| `y-MM`    | `5-03`    | `0005-03-01T00:00:00.000000` | acts as `yy`                                 |
```
act as 'yy' ? 
I doesn't add 2000 to year.

bluestreak01: yeah, it does

bziobrowski: In the line I pasted above it doesn't add 2k to year, while for input containing 2-digit year it does .

amyshwang: I simplified the explanation after trialing with it. I did not list too many examples because I realise the issue is just around `y` and it can be explained using a couple of rules. Hope it makes sense.

bziobrowski (approved):

PR #859, opened 2022-05-24
fix scroll position when clicking on "openings"

argshook:
On `questdb.io/careers` page, the "openings" button scrolls the page
down to align current job openings to the top.

This PR slightly adjusts this position, so that the first opening is not
covered by the navigation bar


jerrinot (approved):
I am not in a position to review the impl, but the staging deployment does what's supposed to do 🎉 thank you!

PR #875, opened 2022-06-02
Fixed a timestamp issue, added instructions to solve dependencies inc…

javier:
…luding M1, changed API rate

* I changed docker compose to start as a daemon so we don't block a terminal only for that. I changed the instructions to check when the cluster is ready by checking docker ps and the status of the connect container
* I found some dependencies issues on an Apple M1, but this might happen also in other systems when installing confluent Kafka. I linked to the github issue where I found my solution and explained the steps to make this work on an M1
* The create table statement had the "timestamp" name with double quotes, and that was causing an error as QuestDB would add the double quotes to the column name. I changed them to backticks and now it works
* I changed the API Rate on the python script (it should be changed on the other repo as well) so now we insert every 15 seconds. I changed that across the rest of the article
* I noticed a typo on the variable name "inteval" of the script and renamed to "interval". I can see that was already fixed on the Soot3 repo :)


pswu11:
Gonna merge once we get a 👍🏼 from the original author

PR #891, opened 2022-06-13
chore(docs): add Golang client to ILP clients list

puzpuzpuz:


docs/develop/insert-data.md
```diff
@@ -97,18 +96,25 @@ Telegraf agent to collect and send metrics to QuestDB via ILP.
...
+  [https://github.com/questdb/go-questdb-client](https://github.com/questdb/go-questdb-client)
+
+- For other languages, we have examples and a
+  [protocol reference](/docs/reference/api/ilp/overview).
 
 ### Examples
 
 These examples send a few rows of input. These use client libraries for C++, C#,
-Java and C, and raw TCP socket connections for NodeJS, Go and Python.
+Golang, Java and C, and raw TCP socket connections for NodeJS and Python.
```
amyshwang, line 117: I'd suggest adding Oxford comma and reduce the use of "and".

How about:

Golang, Java, and C, as well as raw TCP socket connections for NodeJS and Python.

docs/operations/backup.md
```diff
@@ -22,7 +22,7 @@ QuestDB provides two strategies for creating backups:
 There are some limitations of creating backups that users should note about the
 types of supported filesystems and what data is included in backups:
 
-- QuestDB officially supports **EXT4** or **XTS** or any filesystem that
+- QuestDB officially supports **EXT4** or **XFS** or any filesystem that
```
amyshwang, line 25: This is intentional? Does not seem to be relevant to Golang?

amyshwang (changes requested):
I have made a small editorial changes. And some questions:
* I noted some md syntax changes that does not impact the output. Are these changes made to make the content more consistent? (e.g. "-" for bullet point instead of "*")
* Go and Golang - I gather that they are the same. Do we want to choose one? Currently the text uses Goland but the snippets use Go

Thank you!

docs/develop/insert-data.md
puzpuzpuz: Makes sense. Addressed in 5ae2df2072eccc459ec4046bae9d00d3b3cb0182

docs/operations/backup.md
puzpuzpuz: It's a typo fix. There is no such file system as XTS, so it must be XFS.

PR #932, opened 2022-07-11
add Central Group case study page

argshook:
This PR adds a new page with Central Group case study


src/pages/case-study/central-group.tsx
```diff
@@ -0,0 +1,349 @@
...
+            QuestDB is the core engine driving real-time analytics data for
+            Central Group, the largest retail company in Asia.
+          </p>
+        </div>
+
+        <div className={juCss.jumbotron__banner}>
+          <img
+            alt="Logo of Central Group"
+            height={475}
+            src="/img/pages/customers/logos/central_group.svg"
```
amyshwang, line 122: * I don't see an svg file being added in this PR, but it is in the logos folder?
* Is this the missing header image you mentioned? If so, we could use central-of-life.png here (instead of in L195), provided that caption for header is also allowed.

image

amyshwang (commented):
Made suggestion for header image.

For the other missing elements, I will reach out to them and follow up on quotation - the entry block seems to use header image and quote.

src/pages/case-study/central-group.tsx
```diff
@@ -0,0 +1,349 @@
...
+          </p>
+
+          <ul>
+            <li>E-commerce platforms: Online order tracking.</li>
+            <li>Analytical data: Keyword searches, user data analysis.</li>
+            <li>
+              Online marketing data: Unser interaction with social media feeds,
+              online advertisement.
+            </li>
+          </ul>
```
amyshwang, line 268: Bullet list font:
These are smaller than the main body but these are important info, could we enlarge the font at all? Same for Line 329-41.

src/pages/case-study/central-group.tsx
argshook: > I don't see an svg file being added in this PR, but it is in the logos folder?

yes, we have this logo from earlier, so no need to add a new one

> Is this the missing header image you mentioned? If so, we could use central-of-life.png here (instead of in L195), provided that caption for header is also allowed.

yep, for now i just added the logo in place for future header image.
I'll move `central-of-life.png` here

src/pages/case-study/central-group.tsx
```diff
@@ -249,17 +250,16 @@ const CentralGroup = () => {
...
-            <li>Analytical data: Keyword searches, user data analysis.</li>
-            <li>
-              Online marketing data: Unser interaction with social media feeds,
-              online advertisement.
-            </li>
-          </ul>
-
           <p className="font-size--large">
             The CDCs are then streamed into Kafka. QuestDB&#39;s engine consumes
             Kafka topics in real time and acts as long term storage of time
```
amyshwang, line 265: From the output, it seems that this paragraph has a smaller font - but I can see that the setting is the same?
image

src/pages/case-study/central-group.tsx
```diff
@@ -249,17 +250,16 @@ const CentralGroup = () => {
           <p className="font-size--large">
             We deploy a dedicated Change Data Captures (CDCs) process to collect
             endpoints from various source:
+            <ul>
```
amyshwang, line 253: Could we insert an empty line here?

src/pages/case-study/central-group.tsx
```diff
@@ -316,23 +316,22 @@ const CentralGroup = () => {
 
           <p className="font-size--large">
             In the end, QuestDB stood out for the following reasons:
+            <ul>
```
amyshwang, line 319: Same as above - could we insert an empty line here?

amyshwang (commented):
Some minor comments - I am not sure if my reading of the font size is correct...

amyshwang:
1. Header image set.
2. Quote - could we do without?
3. Entry block: could we use the header plus the summary wording - 

```
QuestDB is the core engine driving real-time analytics data for Central Group, the largest retail company in Asia. 

```

amyshwang:
Checked the output - looks good! Let's merge it!

amyshwang (approved):
Thanks!

PR #940, opened 2022-07-14
feat: move tutorials to blog

pswu11:
- [x] move tutorials to blog
- [x] set the url redirection 
- [x] update navbar url


pswu11:
@argshook Tested with two older articles. The redirection seems to work without problem.

argshook (approved):

PR #941, opened 2022-07-14
Unify cloud contact

insmac:
- Refactored email submission form on top of the Cloud page so that it uses `<ContactFormDialog>` component.
- Split components inside `pages/cloud.tsx` into files (and created `modules` directory for that purpose at fist).


argshook (approved):
:+1:

Issue #953, opened 2022-07-20
New features to COPY to be documented


amyshwang:
Location:

https://questdb.io/docs/reference/sql/copy

Source:
https://github.com/questdb/questdb/pull/2155

Task:
1. A small tutorial on how to use `COPY` to import csv files.
2. Expand the current reference page to include the new PR. 
3. Adding a link to the tutorial from the COPY ref page.

PR #955, opened 2022-07-21
chore: adding new `COPY` options to reference #953

amyshwang:
Following changes added in https://github.com/questdb/questdb/pull/2155, this PR update the sql reference and the syntax diagram to reflect the changes.


docs/reference/sql/copy.md
```diff
@@ -20,10 +20,16 @@ is automatic.
...
-  header.
-- `with header false` - QuestDB will use schema recognition to determine
-  whether the first rows should be used as header.
+- `WITH HEADER true/false`: When `true`, QuestDB automatically assumes the first row is a
+  header. Otherwise, schema recognition is used to determine whether the first row is used as header.
+- `PARALLEL`: When enabled, parallel import is on. By default this option is disabled and sequential import is used.
+- `TIMESTAMP`: Define the name of the timestamp column in the file to be imported.
+- `FORMAT`: Timestamp column format when the format is not standard or cannot be detected.
+- `PARTITION BY`: Partition unit to use when target table does not exist.
+- `On ERROR`: Define responses to data parsing errors. The valid values are:
```
puzpuzpuz, line 29:
```suggestion
- `ON ERROR`: Define responses to data parsing errors. The valid values are:
```

puzpuzpuz (commented):
We should also document `COPY tab CANCEL;` command which cancels the in-flight import.

docs/reference/sql/copy.md
```diff
@@ -20,10 +20,16 @@ is automatic.
 
 ## Options
 
-- `with header true` - QuestDB will automatically assume the first row is a
-  header.
-- `with header false` - QuestDB will use schema recognition to determine
-  whether the first rows should be used as header.
+- `WITH HEADER true/false`: When `true`, QuestDB automatically assumes the first row is a
+  header. Otherwise, schema recognition is used to determine whether the first row is used as header.
+- `PARALLEL`: When enabled, parallel import is on. By default this option is disabled and sequential import is used.
```
puzpuzpuz, line 25: We should also document that only a single parallel import is allowed to be executed, i.e. users can't run multiple imports to be executed concurrently.

amyshwang: Also to add these thanks to @bziobrowski 

- Old sequential import is synchronous and returns (in web console or sql) once import is done 
- New import is async, it returns 'ok' message after receiving the command but then it queues it and runs in background
- only one such import can run at a time because it's supposed to use all available server resources

puzpuzpuz: Yes and we need to document all of these aspects properly, including what's written into the `sys.parallel_text_import_log` table.

puzpuzpuz: One more thing to add: we may decide that we want to align old and parallel COPY behavior (although, it'll be a breaking change) - ideally the exact implementation should be chosen automatically, so that the user doesn't have to specify `with parallel` flag; plus, both of them should become async (hence, write their status into the log table). Let me post an update here once the decision is made.

amyshwang: @puzpuzpuz I added a new commit for things I understand but I don't understand the bit on `sys.parallel_text_import_log` :)

puzpuzpuz: It's the table that accumulates all import status updates.

> Let me post an update here once the decision is made.

Yup, we'll be merging COPY implementations into a single one. So, let's wait until things get more stable and then update this PR. Does it make sense?

static/img/docs/diagrams/.railroad
```diff
@@ -48,7 +48,8 @@ case
   ::= 'CASE' ('WHEN' condition 'THEN' value)* ( | 'ELSE' value ) 'END'
 
 Copy
-  ::= 'COPY' tableName 'FROM' fileName  (| 'WITH' 'HEADER' (true|false) ) ';'
```
bziobrowski, line 51: I think PARALLEL was removed .

docs/reference/sql/copy.md
```diff
@@ -37,10 +69,23 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
+security feature preventing random file access by QuestDB.
+
+## Options
+
+- `WITH HEADER true/false`: When `true`, QuestDB automatically assumes the first row is a
+  header. Otherwise, schema recognition is used to determine whether the first row is used as header.
+- `TIMESTAMP`: Define the name of the timestamp column in the file to be imported. This is required for parallel import when the table does not exist.
+- `FORMAT`: Timestamp column format when the format is not standard or cannot be detected.
+- `PARTITION BY`: Partition unit. This is required for parallel import when the table does not exist, and the value cannot be `NONE` or `null`.
+- `ON ERROR`: Define responses to data parsing errors. The valid values are:
```
bziobrowski, line 84: I think it'd make sense to add description to on error options, e.g. 
SKIP_ROW - skip entire row
SKIP_COLUMN - skip column and use default value (null for nullable types, false for boolean, 0 for other non-nullable types)
ABORT - abort whole import on first error, restoring pre-import table status .

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,54 @@ description: COPY SQL keyword reference documentation.
...
-- If the target table does not exist, then it is created using metadata derived
+  - If the target table does not exist, then it is created using metadata derived
   from the file data.
 
-## Options
+- If the original source table is partitioned, parallel import is used:
+
+  - If the target table exist, it must be empty for the operation to perform successfully.
+  - If the target table does not exist, `TIMESTAMP` or `FORMAT` must be defined and `PARTITION BY` cannot be `NONE`, `null`, or skipped.
+  - The operation takes up all the available resources, and therefore only one parallel import is allowed at a time.
```
bziobrowski, line 26: git ?

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,54 @@ description: COPY SQL keyword reference documentation.
...
+Copies data from a delimited text file into a table. `COPY` has the following import modes:
 
-- If the target table exists, then the data is appended provided the file
+- If the original source table is not partitioned, serial import is used:
+
+  - If the target table exists, then the data is appended provided the file
   structure matches the table.
-- If the target table does not exist, then it is created using metadata derived
+  - If the target table does not exist, then it is created using metadata derived
   from the file data.
```
bziobrowski, line 21: source vs target table  ? 
I think it's better to use 'target table' . 

Logic is a bit more complicated because parallel import is used when :
- target table exists  and is partitioned 
- target table doesn't exist and PARTITION BY is specified (to a valid unit ).

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
 
 ## Description
 
-Copies data from a delimited text file into a table. Column delimiter detection
-is automatic.
+Copies tables from a delimited text file into a table. `COPY` has the following import modes:
```
puzpuzpuz, line 13: nit: "Copies tables into a table" sounds weird. The input file contains lines or rows and they're inserted into a single table.

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
...
-  structure matches the table.
-- If the target table does not exist, then it is created using metadata derived
+- Parallel import, used for copying partitioned tables:
+
+  - If the target table exists and is partitioned, the target table must be empty.
+  - If the target table does not exist, `TIMESTAMP` or `FORMAT` must be defined, and `PARTITION BY` should not be `NONE`, `null`, or skipped.
+
+  ::: note
+
+  Parallel import takes up all the available resources, and therefore only one parallel import is allowed at a time.
```
puzpuzpuz, line 22: Both COPY implementations don't support concurrent imports, so this note is misleading.

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
...
-is automatic.
+Copies tables from a delimited text file into a table. `COPY` has the following import modes:
 
-- If the target table exists, then the data is appended provided the file
-  structure matches the table.
-- If the target table does not exist, then it is created using metadata derived
+- Parallel import, used for copying partitioned tables:
+
+  - If the target table exists and is partitioned, the target table must be empty.
+  - If the target table does not exist, `TIMESTAMP` or `FORMAT` must be defined, and `PARTITION BY` should not be `NONE`, `null`, or skipped.
```
puzpuzpuz, line 18: We don't support `null` value for PARTITION BY, so I'm not sure if we should mention this invalid value.

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
...
+
+`COPY` generates a log table,`sys.text_import_log`, tracking `COPY` operation for the last three days with the following information:
+
+| Column name   | Data type | Notes                                                           |
+|---------------|-----------|-----------------------------------------------------------------|
+| ts            | timestamp | The log event timestamp                                         |
+| id            | string    | Import id                                                       |
+| table         | symbol    | Destination table name                                          |
+| file          | symbol    | The source csv file                                             |
+| phase         | symbol    | Import phase.*  Available only to parallel import               |
```
puzpuzpuz, line 43:
```suggestion
| phase         | symbol    | Import phase.* Available only in intermediate log records of parallel import               |
```

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
...
+- partition_import
+- symbol_table_merge
+- update_symbol_keys
+- build_symbol_index
+- move_partitions
+- attach_partitions
+- analyze_file_structure
+- cleanup
+
+Table message retention is configurable through `cairo.sql.copy.log.retention.days` setting, and is three days by default.
```
puzpuzpuz, line 64:
```suggestion
Log table row retention is configurable through `cairo.sql.copy.log.retention.days` setting, and is three days by default.
```

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
...
+
+`COPY` returns `id` value from `sys.text_import_log` to facilitate debugging.
```
puzpuzpuz, line 66: Why do we mention debugging here? We return the id to allow user to track the import progress and await until import finishes. It has nothing to do with debugging.

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
...
 
 :::note
```
puzpuzpuz, line 68: nit: I'd move this and subsequent paragraph to the top of this section. This way it'd be more apparent for the user that they need to change the config to be able to user imports. Would be also nice to mention explicitly that they have to copy their CSV files into the root dir before running an import.

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
-files located in a directory relative to the `copy directory`. This is a
-security feature to disallow random file access by QuestDB.
+The `copy root directory` can be on a local disk to the server, on a remote disk, or
+a remote filesystem. QuestDB enforces that the tables are only written from
+files located in a directory relative to the `copy root directory`. This is a
+security feature preventing random file access by QuestDB.
+
+## Options
+
+- `WITH HEADER true/false`: When `true`, QuestDB automatically assumes the first row is a
```
puzpuzpuz, line 87: `WITH` keyword has to be placed before the very first option, but necessarily `HEADER true/false`.

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
+  header. Otherwise, schema recognition is used to determine whether the first row is used as header.
+- `TIMESTAMP`: Define the name of the timestamp column in the file to be imported.
+- `FORMAT`: Timestamp column format when the format is not standard or cannot be detected.
+- `PARTITION BY`: Partition unit.
+- `ON ERROR`: Define responses to data parsing errors. The valid values are:
+    - `SKIP_ROW`: Skip the entire row
+    - `SKIP_COLUMN`: Skip column and use the default value (`null` for nullable types, `false` for boolean, `0` for other non-nullable types)
+    - `ABORT`: Abort whole import on first error, and restore the pre-import table status
+- `CANCEL`: `COPY '<id>' CANCEL`: using the import `id` to cancel copying a table.
```
puzpuzpuz, line 96: `CANCEL` is not an option, but another type of `COPY` statement.

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
 
 ## Examples
 
+For more details on parallel import, please also see [Importing data in bulk via CSV](/docs/guides/importing-data/).
```
puzpuzpuz, line 100: The linked section doesn't stand for parallel import, but rather HTTP-based import. That one is serial, but we shouldn't mention this as it's an implementation detail. In general, we should mention HTTP-based import as a more convenient alternative to `COPY` for the case of smaller CSV files.

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
+
 ```questdb-sql title="COPY"
-COPY trades20191223 FROM 'C:\archive\trades\20191223.csv'
+COPY weather FROM 'weather.csv' WITH HEADER true FORMAT 'yyyy-MM-ddTHH:mm:ss.SSSUUUZ' ON ERROR SKIP_ROW;
+```
+
+Returns a hexadecimal import id:
```
puzpuzpuz, line 106: nit: the id format doesn't matter much and is a subject to change in future versions
```suggestion
Returns an import id string:
```

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
+
+|  id string      |
+|-----------------|
+|55ca24e5ba328050 |
+
+
+The log can be accessed by querying:
+
+```questdb-sql
+SELECT * FROM 'sys.text_import_log'
```
puzpuzpuz, line 116: nit: it makes more sense to show how to check logs of a concrete import
```suggestion
SELECT * FROM 'sys.text_import_log' WHERE id = '55ca24e5ba328050';
```

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
 ```
+A sample log table low:
```
puzpuzpuz, line 118: nit: what does "low" mean?

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
+|   2022-08-03T10:40:25.586455Z  |   55ca24e5ba328050  |  weather |   weather.csv  |          |   started  |            |                 |                  |   0       |
+|                                |                     |          |                |          |            |            |                 |                  |           |
+
+
+While it's running, import can be cancelled with:
 
-```questdb-sql title="COPY with header true"
-COPY trades20191223 FROM 'C:\archive\trades\20191223.csv' with header true
+```questdb-sql
+copy '55ca24e5ba328050' cancel;
```
puzpuzpuz, line 129:
```suggestion
COPY '55ca24e5ba328050' CANCEL;
```

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
 ```
+
+Within a few seconds import should stop and message with 'cancelled' status should appear in text_import_log, e.g.  
+
+```questdb-sql
+select * from sys.text_import_log where id = '55ca24e5ba328050' order by ts desc limit 1
```
puzpuzpuz, line 135:
```suggestion
SELECT * FROM 'sys.text_import_log' WHERE id = '55ca24e5ba328050' LIMIT -1;
```

docs/reference/sql/copy.md
puzpuzpuz, line 106: nit: would be nice to say that the above SQL statement starts the import asynchronously and returns the import id string.

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
...
+| Column name   | Data type | Notes                                                           |
+|---------------|-----------|-----------------------------------------------------------------|
+| ts            | timestamp | The log event timestamp                                         |
+| id            | string    | Import id                                                       |
+| table         | symbol    | Destination table name                                          |
+| file          | symbol    | The source csv file                                             |
+| phase         | symbol    | Import phase.*  Available only to parallel import               |
+| status        | symbol    | The event status: started, finished, failed, cancelled          |
+| message       | string    | The error message for when status is failed                     |
+| rows_handled  | long      | The counters for the total number of scanned lines in the file  |
```
puzpuzpuz, line 46: We should mention that this and the next counters are only available in the final log row for the given import.

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
...
-Copies data from a delimited text file into a table. Column delimiter detection
-is automatic.
+Copies tables from a delimited text file into a table. `COPY` has the following import modes:
 
-- If the target table exists, then the data is appended provided the file
-  structure matches the table.
-- If the target table does not exist, then it is created using metadata derived
+- Parallel import, used for copying partitioned tables:
+
+  - If the target table exists and is partitioned, the target table must be empty.
```
bziobrowski, line 17: FORMAT is not strictly necessary but should be used if file uses format other than - yyyy-MM-ddTHH:mm:ss.SSSUUUz .

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
-The `copy directory` can be on a local disk to the server, on a remote disk, or
-a remote filesystem. QuestDB will enforce that the tables are only written from
-files located in a directory relative to the `copy directory`. This is a
-security feature to disallow random file access by QuestDB.
+The `copy root directory` can be on a local disk to the server, on a remote disk, or
+a remote filesystem. QuestDB enforces that the tables are only written from
+files located in a directory relative to the `copy root directory`. This is a
+security feature preventing random file access by QuestDB.
+
+## Options
```
puzpuzpuz, line 85: We should describe the default values for all options.

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
+For more details on parallel import, please also see [Importing data in bulk via CSV](/docs/guides/importing-data/).
+
 ```questdb-sql title="COPY"
-COPY trades20191223 FROM 'C:\archive\trades\20191223.csv'
+COPY weather FROM 'weather.csv' WITH HEADER true FORMAT 'yyyy-MM-ddTHH:mm:ss.SSSUUUZ' ON ERROR SKIP_ROW;
+```
+
+Returns a hexadecimal import id:
+
+|  id string      |
```
puzpuzpuz, line 108: nit: no need to include the type here
```suggestion
|  id       |
```

docs/reference/sql/copy.md
```diff
@@ -10,20 +10,60 @@ description: COPY SQL keyword reference documentation.
...
-- If the target table exists, then the data is appended provided the file
-  structure matches the table.
-- If the target table does not exist, then it is created using metadata derived
+- Parallel import, used for copying partitioned tables:
+
+  - If the target table exists and is partitioned, the target table must be empty.
+  - If the target table does not exist, `TIMESTAMP` or `FORMAT` must be defined, and `PARTITION BY` should not be `NONE`, `null`, or skipped.
+
+  ::: note
+
```
bziobrowski, line 21: Should we mention that while one import is running new request(s) will be rejected or show that in tutorial ?

docs/reference/sql/copy.md
```diff
@@ -37,17 +77,64 @@ cairo.sql.copy.root=/Users/UserName/Desktop
...
-a remote filesystem. QuestDB will enforce that the tables are only written from
-files located in a directory relative to the `copy directory`. This is a
-security feature to disallow random file access by QuestDB.
+The `copy root directory` can be on a local disk to the server, on a remote disk, or
+a remote filesystem. QuestDB enforces that the tables are only written from
+files located in a directory relative to the `copy root directory`. This is a
+security feature preventing random file access by QuestDB.
+
+## Options
+
```
bziobrowski, line 86: "WITH" is not part of HEADER option .
WITH behaves as a obligatory 'separator' between copy X from 'source.csv' and options . 
(in PostgreSQL it's optional but let's leave it as is) .

docs/reference/sql/copy.md
puzpuzpuz, line 68: We should also mention `cairo.sql.copy.work.root` directory in this section.

docs/reference/sql/copy.md
amyshwang: But if `TIMESTAMP` is absent and `FORMAT` is not `- yyyy-MM-ddTHH:mm:ss.SSSUUUz` then one of them needs to be defined, right?

docs/reference/sql/copy.md
bziobrowski: The message returned by qdb is  :

"Another import request is in progress. [activeImportId=4650934181329334916]"

it's a bit inconsistent with the table so we should fix it .

docs/reference/sql/copy.md
bziobrowski: ok, so that'sa difference compared to old implementation which allowed loading multiple tables in parallel (with multiple copy commands) .

docs/reference/sql/copy.md
bziobrowski: I think Amy meant troubleshooting but unless we add a whole section on it it's better to use progress tracking.

docs/reference/sql/copy.md
puzpuzpuz: > Should we mention that while one import is running new request(s) will be rejected or show that in tutorial ?

Yeah, would be nice to mention this.

puzpuzpuz: > it's a bit inconsistent with the table so we should fix it .

I'll submit a patch.

docs/reference/sql/copy.md
puzpuzpuz: Same as https://github.com/questdb/questdb.io/pull/955#discussion_r938530841

docs/reference/sql/copy.md
amyshwang: Updated but not sure if it is correct.

docs/reference/sql/copy.md
amyshwang: Adding it here.

docs/reference/sql/copy.md
amyshwang: We're adding parallel import to the page: https://github.com/questdb/questdb.io/pull/967

docs/reference/sql/copy.md
puzpuzpuz: Would be nice to address this one. @amyshwang do you think current text makes sense to the end user?

docs/reference/sql/copy.md
```diff
@@ -10,44 +10,137 @@ description: COPY SQL keyword reference documentation.
...
+  - If the target table exists and is partitioned, the target table must be empty.
+  - If the target table does not exist, and the timestamp format is not `yyyy-MM-ddTHH:mm:ss.SSSUUUz`, `TIMESTAMP` or `FORMAT` must be defined, and `PARTITION BY` should not be `NONE`, or skipped.
 
-- `with header true` - QuestDB will automatically assume the first row is a
-  header.
-- `with header false` - QuestDB will use schema recognition to determine
-  whether the first rows should be used as header.
+- Serial import, used for copying non-partitioned tables:
+
+  - If the target table exists and is not partitioned. The data is appended provided the file structure matches the table.
```
puzpuzpuz, line 22: Should be one sentence:
```suggestion
  - If the target table exists and is not partitioned, the data is appended provided the file structure matches the table.
```

docs/reference/sql/copy.md
```diff
@@ -10,44 +10,137 @@ description: COPY SQL keyword reference documentation.
...
 
-- If the target table exists, then the data is appended provided the file
-  structure matches the table.
-- If the target table does not exist, then it is created using metadata derived
-  from the file data.
+- Parallel import, used for copying partitioned tables:
 
-## Options
+  - If the target table exists and is partitioned, the target table must be empty.
+  - If the target table does not exist, and the timestamp format is not `yyyy-MM-ddTHH:mm:ss.SSSUUUz`, `TIMESTAMP` or `FORMAT` must be defined, and `PARTITION BY` should not be `NONE`, or skipped.
```
puzpuzpuz, line 18: > `TIMESTAMP` or `FORMAT` must be defined

In fact, we only require both to be specified in the described scenario. I'd propose simplifying this sentence to:
```
If the target table does not exist, both `TIMESTAMP` and `PARTITION BY` options must be defined to create a partitioned table and use parallel import. The `PARTITION BY` value should not be `NONE`.
```

docs/reference/sql/copy.md
```diff
@@ -10,44 +10,137 @@ description: COPY SQL keyword reference documentation.
...
+
+:::
+
+`COPY '<id>' CANCEL` cancels the copying operation defined by the import `id`, while an import is taking place. 
+
+### Root directory
+
+`COPY` requires a defined root directory where CSV files are saved and copied from. A CSV file must be saved to the root directory before starting the `COPY` operation. There are two root directories to be defined:
+
+- `cairo.sql.copy.work.root` is used for storing temporary files like indexes or temporary partitions.
```
puzpuzpuz, line 38: We should mention that this property is not mandatory since it defaults to the same value as `cairo.sql.copy.root`. So, only `cairo.sql.copy.root` is a must-have for COPY to be functional.

docs/reference/sql/copy.md
puzpuzpuz: Here is the list of options with default values:
* `header` - `false`
* `format` - `'yyyy-MM-ddTHH:mm:ss.SSSUUUZ'`
* `delimiter` - `','`

docs/reference/sql/copy.md
puzpuzpuz: I didn't see that PR when writing this comment. Resolving it now.

docs/reference/sql/copy.md
amyshwang: Thanks - removing "use parallel import" from your suggestion as the bullet point talks about parallel import and we can make it even shorter.

docs/reference/sql/copy.md
amyshwang: I was thinking about it and forgot to make change! I was debating whether to write:
```
Imports tables from a delimited text file into a table
```

It makes more sense but is far away from the key word `COPY`.

I will try:
```
Copies tables from a delimited text file saved in the defined root directory into QuestDB.
```

puzpuzpuz (approved):

PR #945, opened 2022-07-18
feat: Docs for Java client

jerrinot:
besides adding content for Java ILP client it's also adding a new category to the left menu:
image

The idea is that eventually each client should have its own page in the main documentation site.

PR #958, opened 2022-07-22
remove empty `ilp-client-libraries` file

argshook:
This PR removes an empty file in `docs/reference/ilp-client-libraries.md`.

It is not needed, because after #945 https://questdb.io/docs/reference/ilp-client-libraries/ tries to redirect to https://questdb.io/docs/reference/clients/overview, as specified in [netlify.toml](https://github.com/questdb/questdb.io/pull/945/files#diff-ab8f79b68b7adff7a07db953bf453f3c5aa6ade98d2b1b67d8432b36392489edR170-R173)

However, because empty file exists, the redirect doesn't work and
instead an empty page is shown.

Removing that file fixes the issue and redirect works as expected


jerrinot:
good catch, sorry about that!

amyshwang (approved):
Thanks!

PR #963, opened 2022-07-27
docs: add document for 'substring' function

HeZean:
ref: questdb/questdb#2251


docs/reference/function/text.md
```diff
@@ -181,6 +181,46 @@ FROM example_table;
...
+- `string` is a string to extract from.
+- `start` is an integer specifying the position of the first character to be
+  extracted. Positions start from `1`.
+-  `length` is an integer specifying the count of characters to be extracted.
+  Should be non-negative.
+
+**Return value:**
+
+Returns a string with the extracted characters. If any for the arguments is `null`,
+returns `null`.
```
amyshwang, line 199:
```suggestion
Returns a string with the extracted characters. If any part the arguments is `null`,
the function returns `null`.
```

docs/reference/function/text.md
```diff
@@ -181,6 +181,46 @@ FROM example_table;
...
+```
+
+| id              | country |
+| --------------- | ------- |
+| UK2022072619373 | UK      |
+| UK2022072703162 | UK      |
+| US2022072676246 | US      |
+
+If the `start` argument is negative, the substring will stop at position
+`length + start` if the end position is greater than 1, otherwise empty string.
```
amyshwang, line 214: -5+9=4 but the output stops at position 3? I may have missed something but having played around I observes the  behaviours:

```suggestion
If the `start` argument is negative, the output depends on the value of `start+length`:
- If `start+length` is greater than 1, the substring stops at position `start+length - 1`.
- If `start+length` is zero, the output is empty string.
- If `start+length` is less than zero, the output is `null`.
```
And in this way the example makes sense.

amyshwang (changes requested):
Thank you for the contribution! I have a couple of comments and I hope it makes sense!

HeZean:
@amyshwang thanks a lot, that did make the doc easier to understand

docs/reference/function/text.md
HeZean: well, characters here are 1-indexed

HeZean:
Hi @amyshwang, could you please help to check the CI failings? I checked the log but didn't get the point.

amyshwang:
It is probably a known problem in our checks. Let me re-trigger.

amyshwang (approved):
Thank you for your contribution :)

PR #966, opened 2022-08-02
docs: add examples ILP authentication

jerrinot:
Also: Moved the nodejs example to the last tab,
because the example does not use the nodejs client
and the code with row sockets is long and scary.


docs/reference/api/ilp/authenticate.md
```diff
@@ -323,6 +324,41 @@ if __name__ == '__main__':
 ```    
 </TabItem>
 
+<TabItem value="java">
+
```
amyshwang, line 328: Maybe adding the URL to the Java github here as a comment, like the one in Python?

docs/reference/api/ilp/authenticate.md
```diff
@@ -323,6 +324,41 @@ if __name__ == '__main__':
...
+        // replace:
+        // 1. "localhost:9000" with a host and port of your QuestDB server
+        // 2. "testUser1" with KID from your JSON Web Key
+        // 3. token with the D portion of your JSON Web Key
+        try (Sender sender = Sender.builder()
+                .address("localhost:9009")
+                .enableAuth("testUser1").authToken("GwBXoGG5c6NoUTLXnzMxw_uNiVa8PKobzx5EiuylMW0")
+                .build()) {
+            sender.table("inventors")
+                    .symbol("born", "Austrian Empire")
```
amyshwang, line 346: I don't understand enough to tell but this seems to be very different from the examples in other language. Does it matter? Other language does not use this dummy names?

amyshwang (changes requested):

docs/reference/api/ilp/authenticate.md
jerrinot: I will have at the other languages and will make it consistent. thanks for catching it!

docs/reference/api/ilp/authenticate.md
jerrinot: I think it makes sort of sense for Python, because https://github.com/questdb/py-questdb-client is a repo with Python client and Python client only. readme of the repo has instructions specific to Python client. There is nothing else. 

But https://github.com/questdb/questdb is both server and a client. Client is only a small part for the repo. So I don't think linking from this place makes much sense. What would be the use-case?

amyshwang: You're right - maybe a link to /docs/reference/clients/java_ilp?

plugins/remote-repo-example/index.js
```diff
@@ -10,10 +10,10 @@ const yaml = require("js-yaml")
 
 // "<org>/<repo>/<branch>"
 const sources = [
-  "questdb/c-questdb-client/main",
-  "questdb/py-questdb-client/main",
-  "questdb/go-questdb-client/main",
-  "questdb/nodejs-questdb-client/main",
+  "questdb/c-questdb-client/auth-example",
```
jerrinot, line 13: this is just temporary, to see the effect before corresponding PRs are merged

amyshwang:
SHould I close this for now?

jerrinot:
closing a stale PR, will send a new one

Issue #964, opened 2022-07-28
Outdated info on the last() aggregation


jerrinot:
**Describe the requested changes:**
The [docs says](https://questdb.io/docs/reference/function/aggregation/#last) that the last aggregations work on symbol types only. 

![image](https://user-images.githubusercontent.com/158619/181450711-ad96f7a4-f4a1-4da3-8768-25fd9b443dd5.png)

But there is an implementation for other types too: 
![image](https://user-images.githubusercontent.com/158619/181451008-e5fd9330-95ce-4efd-8a2b-b56ca5deeb1e.png)

The documentation should be updated.

**Location:**
https://questdb.io/docs/reference/function/aggregation/#last

Kudos to Stefan S. for [reporting this](https://questdb.slack.com/archives/C1NFJEER0/p1658993636723769).

jerrinot:
it could be nice also to describe behaviour on a table without a designated timestamp. 
See the discussion: https://questdb.slack.com/archives/C1NFJEER0/p1658995082932059?thread_ts=1658993636.723769&cid=C1NFJEER0

amyshwang:
Fixed by #973

PR #973, opened 2022-08-09
fix: amending supported datatype to first and last #964

amyshwang:


docs/reference/function/aggregation.md
```diff
@@ -203,7 +203,7 @@ FROM (SELECT rnd_double() a FROM long_sequence(100));
 
 ## last
 
-`last(SYMBOL)` - returns the last value of a `SYMBOL` column.
+`last(culomn_name)` - returns the last value of a column. Supported column datatype: `double`, `floating point`, `integer`, `character`, `short`, `byte`, `timestamp`, `date`, `long`, `geohash`.
```
puzpuzpuz, line 206:
```suggestion
`last(column_name)` - returns the last value of a column. Supported column datatype: `double`, `float`, `integer`, `character`, `short`, `byte`, `timestamp`, `date`, `long`, `geohash`.
```

docs/reference/function/aggregation.md
```diff
@@ -119,30 +119,63 @@ SELECT payment_type, count_distinct(counterparty) FROM transactions;
...
+
+```questdb-sql
+SELECT last(device_id) FROM sensors;
+```
+
+|last      |
+|:---------|
+|arduino-03|
+
+Without selecting a designated timestamp column, the table may be unordered:
```
puzpuzpuz, line 170: nit: I'd change the text to "Let's consider an unordered table with the following contents:". The current one doesn't clearly explain that the subsequent queries will run on this new table. Or, alternatively, we could use a different table name, e.g. `unordered_sensors`.

amyshwang: I have added more text - I still keep the wording around designated ts because from that to unordered table might be too much a jump!

docs/reference/function/aggregation.md
```diff
@@ -119,30 +119,35 @@ SELECT payment_type, count_distinct(counterparty) FROM transactions;
...
+`null` values are not counted in `count_distinct` functions.
 
 :::
 
-## first
+## first/last
 
-`first(SYMBOL)` - returns the first value of a `SYMBOL` column.
+- `first(column_name)` - returns the first value of a column. 
+-   `last(column_name)` - returns the last value of a column.
```
puzpuzpuz, line 129: nit:
```suggestion
- `last(column_name)` - returns the last value of a column. 
```

docs/reference/function/aggregation.md
```diff
@@ -152,6 +157,46 @@ SELECT first(device_id) FROM sensors;
...
+
+|first     |
+|:---------|
+|arduino-01|
+
+
+The following query returns the last record for the `device_id` column:
+
+```questdb-sql
+SELECT last(device_id) FROM sensors;
```
puzpuzpuz, line 193:
```suggestion
SELECT last(device_id) FROM sensors_unordered;
```

PR #977, opened 2022-08-09
Adding subpage about importing from multiple CSV files to the guides

javier:
Docs!: Added subpage to the guides for importing multiple CSV files. 

The contents are in this commit, but they are not shown yet on the navigation and will need to be linked from the new Import guide at a later point


bluestreak01:
nice 👍

bluestreak01:
lets see where we place this tomorrow

amyshwang:
Adding this as part of the import guide in #995

PR #967, opened 2022-08-03
chore(doc): added section on copy command to "Importing data in bulk …

bziobrowski:
…via CSV" page

PR #980, opened 2022-08-10
chore(docs): import guide improvement

amyshwang:
1. Addressing comments for #967
- Minor fixes
- Where I don't have an answer I added the comment as internal notes


docs/guides/importing-data.md
```diff
@@ -233,8 +234,23 @@ status table. Given an un-ordered CSV file `weather1mil.csv`, when importing, th
...
 
+<!--
+
+TBD Adam added the following:
+
+- Does the whole need to be restarted from scratch?
+- Can it be resumed?
+- On data issues, is it possible to skip and write out bad data to a file so it can be corrected manually and re-inserted, say, via the REST API?
+[Amy: REST API is not suitable for this page though]
+- What happens in a database crash or OS reboot? Will the table be there but partially written? Will the table be absent?
```
amyshwang, line 248: Attempted answer:
No the table is written as a whole after the import has completed.

docs/guides/importing-data.md
```diff
@@ -233,8 +234,23 @@ status table. Given an un-ordered CSV file `weather1mil.csv`, when importing, th
...
 
 
+
 ## FAQ
 
+<!--
+
+TBD Adam added the following:
+
+- Does the whole need to be restarted from scratch?
```
amyshwang, line 244: This really depends on the error and the way the user installs QuestDB?
Attempted answer:
When root directory is set, QuestDB needs to be restarted for COPY to work.

docs/guides/importing-data.md
```diff
@@ -233,8 +234,23 @@ status table. Given an un-ordered CSV file `weather1mil.csv`, when importing, th
...
+- Can it be resumed?
```
amyshwang, line 245: An import operation cannot be paused and resumed.

docs/guides/importing-data.md
```diff
@@ -233,8 +234,23 @@ status table. Given an un-ordered CSV file `weather1mil.csv`, when importing, th
...
+- On data issues, is it possible to skip and write out bad data to a file so it can be corrected manually and re-inserted, say, via the REST API?
```
amyshwang, line 246: 1. The `On ERROR` option defines error handling.
2. Not sure about how to re-inserted the rows.

docs/guides/importing-data.md
```diff
@@ -233,8 +234,23 @@ status table. Given an un-ordered CSV file `weather1mil.csv`, when importing, th
...
+[Amy: REST API is not suitable for this page though]
```
amyshwang, line 247: I misunderstood the question.

docs/guides/importing-data.md
```diff
@@ -233,8 +234,23 @@ status table. Given an un-ordered CSV file `weather1mil.csv`, when importing, th
...
+- What happens in a database crash or OS reboot? Will the table be there but partially written? Will the table be absent?
+- How does one find the log file?
```
amyshwang, line 249: In /questdb/log? I would add this as part of the Track progress section.

amyshwang (commented):
Adding some attempted answers to @amunra's comments migrated into the files.

docs/guides/importing-data.md
amunra: Let's say that I'm importing 1GiB of data and 10 rows are broken half way through.
Do I get a CSV containing those 10 broken rows that I can edit and re-insert later?

docs/guides/importing-data.md
amunra: See above.

docs/guides/importing-data.md
```diff
@@ -50,11 +50,12 @@ concatenate these files before importing into QuestDB.
 
 ### Configure `COPY`
 
-Enable `COPY` and [configure](/docs/reference/configuration/#csv-import) `COPY` directories to suite your server.
```
bziobrowski, line 53: suite -> suit ?

docs/guides/importing-data.md
bziobrowski: I think Adam asked about restarting import . 
ON ERROR clause determines how errors are treated - they can either be ignored (on row/column level) or abort import . 
Errors are logged in all cases .

docs/guides/importing-data.md
bziobrowski: Failed rows are reported in logs with line number (in indexing phase) or file offset (data import phase) .

docs/guides/importing-data.md
bziobrowski: log messages are mentioned in "Import finished but table is (almost) empty" FAQ point .

docs/guides/importing-data.md
bziobrowski: Only if reboot/power loss happens while partitions are being attached then table might be left with incomplete data . 
If failure happens earlier things will be fine .

docs/guides/importing-data.md
bziobrowski: ON ERROR = SKIP_COLUMN then only this field would be left empty . 

Data parsing is way more problematic than that . For example - a single fixed format is expected for date or timestamp column so any row that's a bit off (e.g. missing or shorter seconds field ) will trigger parse error  .

docs/guides/importing-data-rest.md
```diff
@@ -18,6 +18,11 @@ While REST API is convenient for importing small tables, `COPY` is optimized for
 
 :::
 
+
```
bziobrowski, line 21: It's more complex than just size .
Serial import used in REST isn't that slow, it's just that it uses a single thread .
For sorted input it might be fine for even 100GB-big files but for unsorted (& partitioned target table ) - it can cause MASSIVE O3  . 
MASSIVE as in - it either crashes in the middle or takes very long time and writes 1000 times more data than it should (enough to be concerened about SSD wear).

docs/guides/importing-data.md
```diff
@@ -233,8 +234,23 @@ status table. Given an un-ordered CSV file `weather1mil.csv`, when importing, th
...
+- Does the whole need to be restarted from scratch?
+- Can it be resumed?
+- On data issues, is it possible to skip and write out bad data to a file so it can be corrected manually and re-inserted, say, via the REST API?
+[Amy: REST API is not suitable for this page though]
+- What happens in a database crash or OS reboot? Will the table be there but partially written? Will the table be absent?
+- How does one find the log file?
+
+-->
+
+
```
bziobrowski, line 253: @amyshwang I think some sql in FAQ section would use adding some spacing, e.g. above "ALTER TABLE table_name RENAME COLUMN f0 TO ts;"

docs/guides/importing-data.md
amyshwang: To add a link to - https://questdb.io/docs/reference/configuration/#logging

docs/guides/importing-data.md
```diff
@@ -224,37 +235,85 @@ details of import:
...
+```shell
+dd bs=1 count=1000 skip=56 if=input_file.csv 2>/dev/null | head -1 >> errors.csv
+```
+-->
 
 ## FAQ
 
+<details>
+  <summary>What happens in a database crash or OS reboot?" error message</summary>
```
amyshwang, line 279:
```suggestion
  <summary>What happens in a database crash or OS reboot?</summary>
```

docs/guides/importing-data-rest.md
amyshwang: I am removing the internal notes. We could add this kind of consideration to a block post for import real world scenario. Captured in my content planning table.

docs/guides/importing-data.md
amyshwang: I tried to add more content on error log. As there is no good solution to export the errored rows, I added some content in html internal note for future thought. 
At the moment we encourage users to trial on a smaller table to ensure very few error for the final import.

docs/guides/importing-data.md
```diff
@@ -224,37 +235,85 @@ details of import:
 2022-08-03T14:01:42.791789Z I i.q.c.t.CsvFileIndexer finished chunk [chunkLo=11549510915, chunkHi=15399347885, lines=30000011, errors=0]
 ```
 
-Import into non-partitioned tables uses single threaded implementation that reports only start and finish records in
-status table. Given an un-ordered CSV file `weather1mil.csv`, when importing, the log table shows:
```
bziobrowski, line 228: Error also gets logged when ON ERROR ABORT but import stops on first error .

bziobrowski (approved):

docs/guides/importing-data.md
```diff
@@ -224,37 +235,87 @@ details of import:
...
+dd bs=1 count=1000 skip=56 if=input_file.csv 2>/dev/null | head -1 >> errors.csv
+```
+-->
 
 ## FAQ
 
+<details>
+  <summary>What happens in a database crash or OS reboot?</summary>
+<p>
+
```
bziobrowski, line 283: Maybe we could also say something like 'If that happens then please recreate or truncate table and run import again' ?

amyshwang: Added more content!

PR #981, opened 2022-08-10
feat: document partition detach/attach

marregui:
documents https://github.com/questdb/questdb/pull/2280


marregui:
still working on it

marregui:
ready for review

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,143 @@
...
+:::
+
+## Syntax
+
+![Flow chart showing the syntax of ALTER TABLE with DETACH LIST PARTITION keyword](/img/docs/diagrams/alterTableDetachPartition.svg)
+
+## Configuration
+
+The root file system location where QuestDB stores detached partitions is configurable 
+in `server.conf` by means of attribute:
```
amyshwang, line 23:
```suggestion
in `server.conf` by using the attribute:
```

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,143 @@
...
+        └── ts.d    
+    └── 1970-01-02T00.detached
+        ├── ... 
+        └── ts.d    
+```
+
+The operation saves global metadata files `_dmeta` and `_dcv` alongside column related files to 
+be able to attach the partition at a later time. These metadata files represent the state
+of the table at the time of detaching and do not exist in [backups](../../operations/backup.md).
+Partitions lacking these files will still be attachable, however no checks are performed on the
```
amyshwang, line 86:
```suggestion
Partitions without the  metadata files will still be attachable, but no checks are performed on the
```

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,143 @@
...
+metadata beyond data file integrity, which means the table's structure can not change between a
+backup and an attach from backup. 
+
+Detaching a partition however allows for the table's structure to mutate (add/remove columns, 
+but not rename them).
```
amyshwang, line 91:
```suggestion
Detaching a partition, however, allows for the table's structure to mutate (add/remove columns, 
but not rename them).
```

amyshwang: I am not sure why "however" is used here...

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,143 @@
...
+
+It is not possible to detach the latest partition, where the latest data are appended to (QuestDB 
+refers to this partition as *active* partition), trying to detach it will result in an error.
+
+:::caution
+
+Detaching a partition just tells QuestDB that it no longer is accessible for *reads*.
```
amyshwang, line 98:
```suggestion
For QuestDB's engine, detaching a partition means that it is no longer accessible for *reads*.
```

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,143 @@
...
+However if you insert new data under the detached partition's time range, the partition
```
amyshwang, line 99:
```suggestion
However, if you insert new data under the detached partition's time range, the partition
```

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,143 @@
...
+attached. In this case, you can [drop](alter-table-drop-partition.md)
+the existing version and run the attach query again; or you can rename (in the file system)
+the current detached version from `*.detached` to `*.something_else` and then issue the
+detach command again to detach the partition, point at which you can choose which version 
+to rename back to `.attachable` before issuing the attach query.
+
+:::
+
+To attach back one, or more of these partitions, for instance `1970-01-01T18` and `1970-01-02T00`, 
+first we rename the the above file system to:
```
amyshwang, line 111: Same as the other file - could we specify the suffix?

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -5,45 +5,116 @@ description: ATTACH PARTITION SQL keyword reference documentation.
 ---
 
 Attaches one or more partitions to a table. This feature is designed to support
-attaching partitions from a backup or copying and attaching partitions across a
-network between QuestDB instances.
+attaching detached partitions by means of query [ALTER TABLE DETACH PARTITION](alter-table-detach-partition.md),
```
amyshwang, line 8:
```suggestion
attaching detached partitions by means of query [ALTER TABLE DETACH PARTITION](alter-table-detach-partition),
```
Please could we remove all the suffix in hyperlinks for consistency?

amyshwang (changes requested):
Editorial review.

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,143 @@
+---
+title: ALTER TABLE DETACH PARTITION
+sidebar_label: DETACH PARTITION
+description: DETACH PARTITION SQL keyword reference documentation.
+---
+
+Detaches one or more partitions from a table. This feature is designed to support
```
amunra, line 7: What does detaching mean?
* Does the table get deleted?
* Is it kept, but archived?
* Is the format changed?
* Is this to move data from one database (e.g. holding the latest data) and reattach to another database instance (e.g. the archive database) without having to change format?
* Is this to correct data while it's detached?
* More generally, what is the feature for?

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -5,45 +5,116 @@ description: ATTACH PARTITION SQL keyword reference documentation.
...
+), INDEX(id2 capacity 32) TIMESTAMP(ts) PARTITION BY HOUR;
+```
+
+with a total of 10 partitions, and `server.conf` set with `cairo.sql.detached.root=detdb`,
+the following query will [detach](alter-table-detach-partition.md) 
+the oldest five partitions:
+
+```questdb-sql
+-- either:
+ALTER TABLE sensors DETACH PARTITION LIST '1970-01-01T00', '1970-01-01T06', '1970-01-01T12', '1970-01-01T18', '1970-01-02T00';
```
amunra, line 61: Is there a SQL command to list partitions for a table?

amunra: Also, are these hourly partitions in UTC?

docs/reference/sql/alter-table-detach-partition.md
amunra: Now that I'm reading further, this maybe these questions would be better answered as a guide on data backup and migration (which I think - but I'm not sure - this feature fulfills).

amunra: Confusingly the "attach" guide states "Moving partitions between tables or database instances is not supported.". I'm pretty lost now.

amyshwang: Quick notes from discussion with @bluestreak01 

Purpose - Decouple partitions from the QuestDB instance. Content not manipulated. Allowing users to deposit partitions to cold storage.
Constraint - only to the same table.
Detached partition treatment: QuestDB does not do anything - it leaves room for manual intervention.

docs/reference/sql/alter-table-attach-partition.md
amyshwang: - I think it is just a directory so you won't need SQL to show them?
- I am making an educated guess that the partitions are decided by the designated timestamp of the table, but I am not very well-educated so maybe I am talking nonsense...

marregui: hola, no, there are no commands to list partitions per table, nor their status (attached, live, detached), nor to manage them, i.e. rename detached to attachable or move them about. It all needs to be done by hand (in cloud is different).
Timestamps are UTC, these partition names are UTC as well, they are created as data turns up for the specific bucket (defined by the timestamp)

docs/reference/sql/alter-table-detach-partition.md
marregui: The use case, in general, is:
- I have lots of data in my table, and it is partitioned by some granularity that makes me happy (by HOUR say).
- If only I could simply use the last 5 items of said granularity, instead of all my data, i.e. only the last 5 hours...
- oh, wait, I can detach partitions older than 5 hours from now and just focus on "hot data"
- later on I can attach whatever partitions I want if I need more data

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,134 @@ sidebar_label: ATTACH PARTITION
...
+Attaches one or more partitions to the table from where the partitions were previously detached. This feature 
+is designed to support restoring partitions detached by using [ALTER TABLE DETACH PARTITION](alter-table-detach-partition).
+
+## Syntax
+
+![Flow chart showing the syntax of ALTER TABLE with ATTACH PARTITION keyword](/img/docs/diagrams/alterTableAttachPartition.svg)
+
+## Configuration
+
+The root file system location where QuestDB stores detached partitions is configurable
```
bluestreak01, line 16: this paragraph is not relevant for attaching partitions. I would remove it or rephrase it somehow. I would suggest we outline sequence of events first to make it look simpler:

1. Copy partition inside table and make sure partition name has suffix `.attachable`. For example `2020.attachable`
2. Execute SQL `alter table x attach partition list '2020'`. Notice how SQL does not include the suffix.

After SQL is executed, QuestDB will rename partition to its correct name and make it available to reading.

bluestreak01 (changes requested):

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,134 @@ sidebar_label: ATTACH PARTITION
...
-attaching partitions from a backup or copying and attaching partitions across a
-network between QuestDB instances.
+Attaches one or more partitions to the table from where the partitions were previously detached. This feature 
+is designed to support restoring partitions detached by using [ALTER TABLE DETACH PARTITION](alter-table-detach-partition).
+
+## Syntax
+
+![Flow chart showing the syntax of ALTER TABLE with ATTACH PARTITION keyword](/img/docs/diagrams/alterTableAttachPartition.svg)
+
+## Configuration
```
bluestreak01, line 14: I would separate description and configuration, in this section both are mixed up

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,134 @@ sidebar_label: ATTACH PARTITION
...
 detached partitions along with example commands for copying and attaching
 partitions can be found in the [examples section below](#examples).
 
+There is an additional configuration attribute in `server.conf`:
+
+```shell
+cairo.attach.partition.copy=false
+```
+
+which defaults to false. When set to true detached partitions are copied to the *detached storage*
```
bluestreak01, line 44: factually incorrect.

remember that `alter table x attach partition list '2020'` will rename `2020.attachable` to something like `2020.201` where `201` is current transaction number of the table.

when

```
cairo.attach.partition.copy=true
```
is set, instead of renaming directory, questdb will copy it, e.g. after SQL both `2020.attachable` and `2020.201` will be present. It might be needed in some niche cases.

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,134 @@ sidebar_label: ATTACH PARTITION
...
+instead of `linked`.
+
 :::caution
 
-This operation is only supported on the partiton was copied from the same table at some point in the past.
-Moving partitions between tables or database instances is not supported.
+This operation is only supported if the partition was detached, or has been copied
```
bluestreak01, line 49: i would also add that operation will also fail when partition already exists, which might happen when someone inserted out of order records into "detached" partition

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,134 @@ sidebar_label: ATTACH PARTITION
...
+```shell
+cairo.detach.root=db
+```
+By default detached partitions are stored in the same folder as the data folder `db`. 
 
 Before running an `ATTACH PARTITION` query, partitions must first be copied in
-the format `<partition_name>.detached` to the storage directory of the table to
-which they will be attached. Given a partition named `2020`, it must be copied
-to the destination table as `2020.detached`. Details of the naming format for
+the format `<partition_name>.SUFFIX` to the *detached storage* directory of
```
bluestreak01, line 25: I would remove this paragraph. It is factually incorrect. The partition must be copied not to *detached storage* but rather inside the table

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,134 @@ sidebar_label: ATTACH PARTITION
...
-This operation is only supported on the partiton was copied from the same table at some point in the past.
-Moving partitions between tables or database instances is not supported.
+This operation is only supported if the partition was detached, or has been copied
+from the same table at some point in the past. Moving partitions between tables or 
+database instances is not supported.
 
 :::
 
-## Syntax
+## Examples
```
bluestreak01, line 55: Before that I would create configuration section and add link to the relevant configuration properties and their description

docs/reference/configuration.md
```diff
@@ -336,7 +339,7 @@ QuestDB.
...
 | cairo.sql.column.purge.retry.limit.days        | 31                | Number of days purge system will continue to re-try deleting stale column files before giving up.                                                                                                                        |
 | cairo.system.table.prefix                      | sys.              | Prefix of the tables used for QuestDB internal data storage. These tables are hidden from QuestDB web console.                                                                                                           |
-| cairo.sql.detached.root                        | db                | Detached storage directory for `ALTER TABLE ATTACH PARTITION` and `ALTER TABLE DETACH PARTITION`          |
+| cairo.sql.detached.root                        | db                | Detached storage directory for `ALTER TABLE ATTACH PARTITION` and `ALTER TABLE DETACH PARTITION`                                                                                                                         |
```
amyshwang, line 342: I figure that this has been removed and changed to `cairo.detach.root`?

docs/reference/configuration.md
```diff
@@ -264,11 +264,14 @@ QuestDB.
...
 | cairo.snapshot.instance.id                     | empty string      | Instance id to be included into disk snapshots.                                                                                                                                                                          |
 | cairo.snapshot.recovery.enabled                | true              | When `false`, disables snapshot recovery on database start.                                                                                                                                                              |
+| cairo.detach.root                              | db                | Directory for storing detached table partitions and metadata. This directory is inside the server root directory provided at startup.                                                                                    |
+| cairo.attach.partition.suffix                  | .attachable       | Suffix in `<partition-name><suffix>`, e.g. 2022-08-30.attachable, to identify a detached partition directory before performing an attach.                                                                                |
```
amyshwang, line 268: This is a background setting that I don't think that a user needs to reset - do we really need it here?

docs/reference/configuration.md
```diff
@@ -264,11 +264,14 @@ QuestDB.
...
+| cairo.attach.partition.copy                    | false             | When `true` detached partition folders are copied to the location defiled by `cairo.detach.root` (instead of being linked).                                                                                              |
```
amyshwang, line 269: I am not sure if I understand. Vlad has some concerns about the content added to this setting below...

docs/reference/sql/alter-table-attach-partition.md
amyshwang: We agree that this is not for general users so removed.

docs/reference/sql/alter-table-attach-partition.md
amyshwang: Config now removed.

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,91 @@ sidebar_label: ATTACH PARTITION
...
 
-![Flow chart showing the syntax of the ALTER TABLE keyword](/img/docs/diagrams/alterTable.svg)
-![Flow chart showing the syntax of ALTER TABLE with ATTACH PARTITION keyword](/img/docs/diagrams/alterTableAttachPartition.svg)
+### Example 1
 
-## Examples
+Given a table `sensors` that has previously [detach](alter-table-detach-partition.md) partitions, if we want to attach one or more of these partitions back, for instance `1970-01-01T18` and `1970-01-02T00`:
```
bluestreak01, line 40:
```suggestion
Given a table `sensors` that has previously [detached](alter-table-detach-partition.md) partitions, if we want to attach one or more of these partitions back, for instance `1970-01-01T18` and `1970-01-02T00`:
```

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,91 @@ sidebar_label: ATTACH PARTITION
...
+    └── 1970-01-02T00.attachable
+        ├── ... 
+        └── ts.d    
+```
+
+2. Then we perform the query:
+
+```questdb-sql
+ALTER TABLE sensors ATTACH PARTITION LIST '1970-01-01T18', '1970-01-02T00';
+--
```
bluestreak01, line 72: why do we have this comment ?

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,91 @@ sidebar_label: ATTACH PARTITION
...
+To attached one or more partitions that were previously detached to the same table:
+
+1. Copy the partition file(s) to the table directory nested in the root directory. The root directory is set by `cairo.root` and is set to `db` by default.
+2. Amend the partition file suffix to `attachable`.
+
+    For example, given a partition named `2020`, it must be copied to the destination 
+    table's detached storage as `2020.attachable`. Details of the naming format for
+    detached partitions along with example commands for copying and attaching
+    partitions can be found in the [examples section below](#examples).
+3. Run the `ALTER TABLE ATTACH PARTITION` command.
```
bluestreak01, line 24: it is a matter of opinion perhaps, but since we asked to rename directory first and then run SQL, should we be specific that partition name reference from SQL should not include `.attachable` suffix?

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -4,46 +4,91 @@ sidebar_label: ATTACH PARTITION
...
+
+2. Then we perform the query:
+
+```questdb-sql
+ALTER TABLE sensors ATTACH PARTITION LIST '1970-01-01T18', '1970-01-02T00';
+-- 
+-- WHERE clauses are not supported for ATTACH queries
+```
+
+### Example 2
```
bluestreak01, line 76: This is really bad example. Did this example work for you?

`DETACH` SQL writes metadata into partition, therefor `ATTACH` only works with `DETACHED` partitions. Partitions from a backup are not `DETACHED` per se and probably won't work

docs/reference/sql/alter-table-attach-partition.md
```diff
@@ -93,8 +137,8 @@ ALTER TABLE sensor_data ATTACH PARTITION LIST '2019';
 Multiple partitions may be copied and attached to a table:
 
 ```bash title="Copying multiple partitions to the sensor_data table"
-cp -r ~/backup/2021-01-01/sensor_data/2020 /path/to/questdb/db/sensor_data/2020.detached
-cp -r ~/backup/2021-01-01/sensor_data/2019 /path/to/questdb/db/sensor_data/2019.detached
+cp -r ~/backup/2021-01-01/sensor_data/2020 /path/to/questdb/db/sensor_data/2020.attachable
+cp -r ~/backup/2021-01-01/sensor_data/2019 /path/to/questdb/db/sensor_data/2019.attachable
 ```
 
 The partitions are attached by providing a comma-separated list of partitions:
```
bluestreak01, line 144: this example is redundant, example 1 shows how to attach multiple partitions

bluestreak01 (changes requested):

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,108 @@
...
+        └── ts.d    
+    ├── 1970-01-01T18.detached
+        ├── ... 
+        └── ts.d    
+    └── 1970-01-02T00.detached
+        ├── ... 
+        └── ts.d    
+```
+
+The operation saves global metadata files `_dmeta` and `_dcv` alongside column-related files to
```
bluestreak01, line 81: This is very heavy. I would not mix-in `backups` here nor divulge details of how `detach` works under the hood.

What we should explain instead is that `.detached` directories can be moved elsewhere. 

I would suggest we do not talk about modifying content of the directories. We should discourage that.

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,108 @@
...
+- For a table to support this feature it has to have a [designated timestamp](../../concept/designated-timestamp) column.
+- QuestDB does not manipulate the detached partitions. It is for the users to decide whether to compress the files and where to migrate them to.
+
+:::
+
+## Syntax
+
+![Flow chart showing the syntax of ALTER TABLE with DETACH LIST PARTITION keyword](/img/docs/diagrams/alterTableDetachPartition.svg)
+
+## Steps
```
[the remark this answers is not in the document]
bluestreak01: Another example, we talk at bottom of the page what detaching means. This can be in the description of the SQL. We also should answer `why` anyone would want to detach partition. Why can't they move partitions without detaching?

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,108 @@
...
+
+Detaching a partition, however, allows for the table's structure to mutate (add/remove columns, 
+but not rename them).
+
+It is not possible to detach the latest partition, where the latest data are appended to (QuestDB 
+refers to this partition as an *active* partition), and attempts to detach it will result in an error.
+
+:::caution
+
+For QuestDB's engine, detaching a partition means that it is no longer accessible for *reads*.
```
bluestreak01, line 96: This note ultimately belongs to `attach` page. The reason is that `DETACH` SQL will never fail, only if table is "busy".

On other hand `ATTACH` may fail because partition could have been recreated by O3

bluestreak01 (changes requested):

docs/reference/sql/alter-table-detach-partition.md
```diff
@@ -0,0 +1,108 @@
...
+Given a table `sensors` defined as:
+
+```questdb-sql
+CREATE TABLE sensors AS (
+    SELECT
+        rnd_symbol('ALPHA', 'OMEGA', 'THETA') id1, 
+        rnd_symbol('PHI', 'SIGMA', 'EPSILOM') id2, 
+        timestamp_sequence(0, 21600000000) ts 
+    FROM long_sequence(10)
+), INDEX(id2 capacity 32) TIMESTAMP(ts) PARTITION BY HOUR;
```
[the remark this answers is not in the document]
amyshwang:
```suggestion
), TIMESTAMP(ts) PARTITION BY HOUR;
```
The existing query works but when I updated it to the above I hit an error "unexpected token"...

docs/reference/sql/alter-table-detach-partition.md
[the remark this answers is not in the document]
amyshwang: why anyone would want to detach partition:

> This feature is designed to allow users to manipulate detached partitions and migrate them to a cold storage. 

Why can't they move partitions without detaching?
My answer would be something like:
Using `ALTER TABLE DETACH PARTITION` ensures relevant metadata is saved with the partition files, and the partitions can be attached back to the same table in the future. Deleting the partition files directly from the table directory means that the detached partitions won't be able to be attached back to the table.

docs/reference/sql/alter-table-detach-partition.md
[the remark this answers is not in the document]
amyshwang: Removed the comma and it worked.

docs/reference/sql/alter-table-detach-partition.md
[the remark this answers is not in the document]
bluestreak01: We could keep the first sentence only. Explaining that detach does something extra is enough to explain “why”

bluestreak01 (approved):

amyshwang (approved):

PR #1007, opened 2022-08-19
chore(docs): fix typo in vacuum table reference

puzpuzpuz:
The typo was `automitically`


docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
 
 When a table is appended Out of Order, a new version of a partition can be
 written to the disk. The old partition version directory is deleted once it is
```
amyshwang, line 10:
```suggestion
When a table is appended in an out-of-order manner, the `VACUUM TABLE` command writes a new partition version to the disk. The old partition version directory is deleted once it is
```

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
-
-The `VACUUM TABLE` command starts a new scan over table partitions directories and
-and column files. It detects redundant, unused files consuming the disk space and deletes them.
-`VACUUM` executes asynchronously e.g. it may keep scanning and deleting files after their response 
-is returned to the SQL client.
-
-This command has to be executed rarely and provides a manual mechanism to reclaim the disk space.
-The implementation scans file system catalogs to detect duplicate directories and files and may be
-relatively expensive to run too often.
+not read by SELECT queries. In the event of file system errors, physical
```
amyshwang, line 11:
```suggestion
not read by `SELECT` queries. In the event of file system errors, physical
```

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+deletion of old files can be interrupted and an outdated partition version may
```
amyshwang, line 12:
```suggestion
deletion of old files may be interrupted and an outdated partition version may
```
I use may because I think this is a possibility, not a capacity (as this is a warning?)

Maybe this should be a note:

```
:::note
In the event of file system errors, physical deletion of old files can be interrupted and an outdated partition version may be left behind consuming the disk space.
:::
```

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+be left behind consuming the disk space.
+
+In a similar manner executing `UPDATE` SQL statement copies column files. Old
```
amyshwang, line 15:
```suggestion
In a similar manner to executing `UPDATE` SQL statement copies column files. Old
```
I don't understand this sentence. It is not complete. Are we trying to say:
Executing `VACUUM TABLE` SQL statement is similar to executing `UPDATE` SQL statement?
Or maybe:
Similar to the `UPDATE` SQL statement, `VACUUM TABL` also copies the column files (of the table?)?

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+column files are automatically deleted but in some limited circumstances can be
```
amyshwang, line 16:
```suggestion
column files are automatically deleted but in certain circumstances, they can be
```

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+left behind. In this case `VACUUM TABLE` will re-trigger delete process of
```
amyshwang, line 17:
```suggestion
left behind. In this case, `VACUUM TABLE` will re-trigger the deletion process of the
```
So reading this my understanding is that if I tun `VACUUM TABLE` for a table with leftover partitions, the command will only do the deletion: it deletes the old files of the table. (And therefore does not write new partition?)

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+updated column files.
+
+The `VACUUM TABLE` command starts a new scan over table partitions directories
```
amyshwang, line 20:
```suggestion
The `VACUUM TABLE` command starts a new scan over table partition directories
```
The main noun here is the directories, right? If so, the above suggestion should work.

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+and and column files. It detects redundant, unused files consuming the disk
+space and deletes them. `VACUUM` executes asynchronously, i.e. it may keep
```
amyshwang, line 22:
```suggestion
space and deletes them. `VACUUM TABLE` executes asynchronously, i.e. it may keep
```

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+scanning and deleting files after their response is returned to the SQL client.
```
amyshwang, line 23: It is not clear what kind of response is expected? Is "SQL client" the right term here? 
I think maybe a normal response should be explained first.
```When executed, `VACUUM TABLE` returns [a message] [...]. As the command executes asynchronously, the process of scanning and deleting may continue after the respond is sent.```

File: docs/reference/sql/vacuum-table.md [hunk 10]
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+
+This command has to be executed rarely and provides a manual mechanism to
```
amyshwang, line 25:
```suggestion
This command has to be executed sparingly and provides a manual mechanism to
```
Do we mean you should not use it too often (sparingly)? 
I also don't understand the second half of the line. Do we mean:
The command provides a manual mechanism to reclaim the disk space.
If so it should be at the start of the intro?

Or maybe it is an explanation to why we should not be using it too often?
Then:
"because it is a manual mechanism..."

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+reclaim the disk space. The implementation scans file system catalogs to detect
```
amyshwang, line 26:
```suggestion
reclaim the disk space. The implementation scans file system catalog to detect
```
I am not sure about this but I think there is only one file system catalog?

docs/reference/sql/vacuum-table.md
```diff
@@ -8,22 +8,24 @@ The `VACUUM TABLE` command triggers partition and column version cleanup.
...
+duplicate directories and files and may be relatively expensive to run too
+often.
```
amyshwang, line 28:
```suggestion
duplicate directories and files and frequent usage of the command can be relatively expensive.
```

amyshwang (changes requested):
I left suggestions but also many questions!

docs/reference/sql/vacuum-table.md
puzpuzpuz: I don't think we should break vacuum description into multiple sections. It's compact and logical. Also, notes are good for some hints while this sentence simply describes why vacuum table may be needed.

docs/reference/sql/vacuum-table.md
puzpuzpuz: Going to change this sentence to make it more logical.

docs/reference/sql/vacuum-table.md
puzpuzpuz: Yes, your understanding is correct.

docs/reference/sql/vacuum-table.md
puzpuzpuz: > Is "SQL client" the right term here?

Yes, to me this term is a correct one.

docs/reference/sql/vacuum-table.md
puzpuzpuz: Going to remove "catalog" to avoid confusion.

docs/reference/sql/vacuum-table.md
puzpuzpuz: Going to restructure this paragraph to make it more logical.

puzpuzpuz:
@amyshwang PTAL

docs/reference/sql/vacuum-table.md
```diff
@@ -6,26 +6,26 @@ description: VACUUM TABLE SQL keyword reference documentation
...
-The `VACUUM TABLE` command starts a new scan over table partitions directories
+When a table is appended in an out-of-order manner, the `VACUUM TABLE` command
+writes a new partition version to the disk. The old partition version directory
+is deleted once it is not read by `SELECT` queries. In the event of file system
+errors, physical deletion of old files may be interrupted and an outdated
+partition version may be left behind consuming the disk space.
+
+When an `UPDATE` SQL statement is run, it copies column files. Old column files
+are automatically deleted but in certain circumstances, they can be left behind.
+In this case, `VACUUM TABLE` will re-trigger the deletion process of the updated
```
amyshwang, line 17: Just for my own understanding perhaps:
So `VACUUM TABLE` is used when `UPDATE` created left old column files behind? And when this happens, users should use `VACUUM TABLE` to tidy up disk space?
The current wording seems to say that you can use `UPDATE` and `VACUUM TABLE` will be triggered (automatically) in certain circumstances.
I am guess that this paragraph is explaining when `VACUUM TABLE` will be useful? I can suggest something once I confirm what it means!

amyshwang (commented):
One question!

docs/reference/sql/vacuum-table.md
puzpuzpuz: Yes, your understanding is correct. Vacuum command is there for those rare cases when the database failed to clean up the old version files due to some circumstances (say, a crash).

amyshwang: But I suppose it is not limited to situation caused by `UPDATE`? 
I'd remove this paragraph from the description, as it is not a general behaviour. It is perhaps more suitable for the example section, to provide more context to the scenario for the command?
```When an UPDATE SQL statement is run, it copies column files. Old column files are automatically deleted but in certain circumstances, they can be left behind. In this case, VACUUM TABLE will re-trigger the deletion process of the updated column files.```

Comment (puzpuzpuz) [reply docs/reference/sql/vacuum-table.md:17]: This paragraph describes the main use case for the vacuum command, so I'd keep it in the description.

Comment (amyshwang) [reply docs/reference/sql/vacuum-table.md:17]:
```suggestion
When an `UPDATE` SQL statement is run, it copies column files of the selected table. The old column files are automatically deleted but in certain circumstances, they can be left behind.
In this case, `VACUUM TABLE` can be used to re-trigger the deletion process of the updated
```

puzpuzpuz:
@amyshwang this one should be good to go.

amyshwang (approved):
LGTM!

PR #1008, opened 2022-08-22
chore(docs) Changing sidebar order and FAQ link for commit lag

amyshwang:


jerrinot:
This is an improvement, but I still have a hard time to understand when exactly will my data become visible. The FAQ is linking to 2 articles: https://deploy-preview-1008--unruffled-blackwell-31bfb2.netlify.app/docs/guides/out-of-order-commit-lag/ and https://deploy-preview-1008--unruffled-blackwell-31bfb2.netlify.app/docs/reference/api/ilp/tcp-receiver/#commit-strategy

The first article describes the concept of "commit lag" (good), while the 2nd article talks about "commit strategies". As a user, I feel these 2 concepts are related, but how exactly they are linked is unclear. For example the 2nd article talks about "idle table timeout" and "interval based commit", while the 1st article does not mention these 2 things at all. I think that's probably due to history: The description of "Commit Strategies" is likely older than the concept of "Commit Lag".  So when the support for Out of order ingestion was introduced then the "Commit Lag" article was created. That's my guess anyway. 

 As a user I would appreciate a flow diagram or something. Something what connects all these concepts and configuration together. One place which can answer a simple question: "I have these parameters when will my data become visible?". 

To be fair I don't know the exact rules myself. There are not simple. But that's a reason to make sure the documentation is super clear. I am happy to help with that! 

My proposal: 
1. Let's merge this. It's still an improvement compared to the current state
2. Try to write an article explaning all concepts one has to know to understand what will data become visible. Happy to help with that, it will be a learning exercise for me too!

jerrinot (approved):
+1 after talking to Amy.

There is a plan for bigger changes, with more examples, etc. but it will likely be influenced by replication (WAL) too.

Issue #965, opened 2022-07-31
Bug - Incorrect timestamps in ISO timestamp string table


salsasepp:
**Bug description:**

The conversion to timestamp values (right column) from ISO 8601 strings (left column) is wrong for ISO strings that contain an UTC offset. This affects three rows in the table referenced below.

Corrected values (right column):

| Valid STRING Format | Resulting Timestamp |
| ----- | ----- |
2010-01-12T12:35:26.123456+01:30 | 2010-01-12T**11**:05:26.123456Z
2010-01-12T12:35:26.123456+01 | 2010-01-12T**11**:35:26.123456Z
[...]
2010-01-12 12:35:26.123456-02:00 | 2010-01-12T**14**:35:26.123456Z

There may be more conversion errors of this type throughout the docs.

See also this recent fix for incorrect timestamp arithmetics: https://github.com/questdb/questdb/commit/78ab4afe712de3a0a87b8a9b5ade5d8814e4d8ef

**Location:**

https://questdb.io/docs/reference/sql/where#native-timestamp-format

**Screenshots:**

**Browser and OS:**

amyshwang:
Fixed by #1012

PR #1012, opened 2022-08-23
fix(docs): fixing wrong timestamp #965

amyshwang:
Checking strong/timestamp conversion in this section


amyshwang:
@salsasepp Thank you for raising this!

jerrinot (approved):
good one!

Issue #988, opened 2022-08-12
Update - Clearer docs for `SYMBOL` and `INDEX` when info covers more than one topic


amyshwang:
**Describe the requested changes:**

1. Feedback from @javier:

If I go to the [Symbol page](https://questdb.io/docs/concept/symbol) I see the info about symbols, but if I want to learn more about [index](https://questdb.io/docs/concept/indexes/)ing I have links both to index and [create table](https://questdb.io/docs/reference/sql/create-table/). I go to index, expecting to find ALL the info about indexes there. But not really. Index tells me a generic overview about indexes in a database, but if I want to really know about index parameters and QuestDB specific info about indexes, I can find it only at the create table page, where index capacity is discussed. Not only that, but on that page I find that there is also the CACHE option for Symbols, which is not related to index, but was not referenced at the Symbols page. Bottomline here is that it is difficult to get the full picture about symbols or indexes, and the only way to do it is on the create table page, which is probably not what I would expect. I have seen this pattern in some other functionalities as well

 2. Cloud user Ventura - Setting partitions and schema with symbols properly is still not well understood
 
**Location:**
TBD

amyshwang:
Another relevant page: - https://deploy-preview-987--unruffled-blackwell-31bfb2.netlify.app/docs/operations/capacity-planning

bziobrowski:
cache - if true then symbol values are cached in java heap (as opposed to mapped file memory) so they should be faster to retrieve 
It doesn't depend on index clause in any way .

bziobrowski:
capacity - refers to index file used for checking symbol uniqueness . If capacity declared  in table definition is much lower than actual (in data) then uniqueness check slows down .

puzpuzpuz:
```sql
CREATE TABLE
  my_table(symb SYMBOL capacity 64 cache index capacity 256,
  price DOUBLE, ts TIMESTAMP, s STRING)
  timestamp(ts)
PARTITION BY DAY;
```

There are two different capacities involved:
1. Symbol capacity (64 in the above example). That's an upper bound for the expected count of distinct symbol values. To make things simpler to understand, think of it as the number of buckets in a hash table - if you set it to a too low value, your lookups might end up with O(n) algorithmic complexity. Symbol capacity is also used to set the initial symbol cache size when the cache is enabled for column.
[...]

PR #1014, opened 2022-08-24
fix(docs): cairo.default.symbol.cache.flag default should be true

amyshwang:


docs/reference/configuration.md
```diff
@@ -268,7 +268,7 @@ QuestDB.
...
 | cairo.create.as.select.retry.count             | 5                 | Number of types table creation or insertion will be attempted.                                                                                                                                                           |
 | cairo.default.map.type                         | fast              | Type of map used. Options: `fast` (speed at the expense of storage), `compact`.                                                                                                                                          |
-| cairo.default.symbol.cache.flag                | false             | When `true`, symbol values will be cached on Java heap.                                                                                                                                                                  |
+| cairo.default.symbol.cache.flag                | true              | When `true`, symbol values will be cached on Java heap.                                                                                                                                                                  |
```
puzpuzpuz, line 271: nit: I'd improve the description to make it more apparent that this setting applies to each table & column individually
```suggestion
| cairo.default.symbol.cache.flag                | true              | When `true`, symbol values will be cached on Java heap instead of being looked up in the database files. Applied to each created table or symbol column, unless not set explicitly via SQL `CACHE` keyword.                                                                                                                                                                  |
```

puzpuzpuz (approved):

docs/reference/configuration.md
amyshwang: My current thoughts are to keep it simple here. Then as part of #988 I should make the global and local setting clear in the page dedicated to `symbol` and index.

There are other settings on this config page has a more local option but we don't specify it here, so I think it is better to leave it out to reduce maintenance cost - wdyt?

puzpuzpuz: I'd prefer to keep config prop descriptions precise. Currently they're somewhat misleading.

amyshwang:
```suggestion
| cairo.default.symbol.cache.flag                | true              | When `true`, symbol values will be cached on Java heap instead of being looked up in the database files.                                                                                                                                                                  |
```

Would this be a good solution? I am leaving the global-local setting out from the table here.

puzpuzpuz: Well, it's better than what we have now, so I'm fine with the suggested change.

amyshwang: Let me commit this. For an overall improvement, let's look at #1017 to decide if wording should be added in this config page too (for now I don't think so but I am open for discussion).

Issue #1013, opened 2022-08-23
Third-party tools


amyshwang:
Location:

![Image](https://user-images.githubusercontent.com/107401731/186206373-6a4e3c7e-93ad-4a2b-a40a-20dc3697cec9.png)

1. Overview page
    - Adding overview to highlight supported tools and "coming soon"
    - 
2. Missing tools to be added:

- [ ] redpanda #1387
- [ ] mindsdb
- [ ] cube.js 
Extract content form tutorials and add it to the sidebar

amyshwang:
### Third-party tools overview page

QuestDB support many third-party tools to meet our users' need. The below is a summary of what we currently support and links to individual guides. We also include a list of tools that we are planning to support.

#### Supported tools
[alphabetical order of the list with links to guide page]
Question: alphabetical order or order based on popularity? 

- [Grafana](url): An observability and monitoring application
- [Kafka Connect](url): A event streaming tool
- [Prometheus](url): An open-source system monitoring and alerting toolkit
- [Telegraf](url): A plugin-driven server agent for collecting & reporting metrics
- 
#### Coming soon
The following are tools that we are planning to support, watch this space!
If your preferred tools are not on this list, please let us know on our [Slack community](link).

[alphabetical order of the list with links to those tools]

amyshwang:
As discussed with @pswu11 - 
1. Try to create an overview page with minimum effort, in order to 
    - Introduce roadmap link
    - Show links to supported tools not having their own individual page yet
This can be an intermediary step, to see if it improves readability.

argshook:
We now have https://questdb.io/docs/third-party-tools

in addition, we can mark sidebar element with custom tags, like `new` or similar.

Thus, closing the ticket.

PR #1018, opened 2022-08-26
chore(docs): Third-part tools overview page (#1013)

amyshwang:


docs/third-party-tools/overview.md
```diff
@@ -0,0 +1,20 @@
...
+We have a growing number of supported tools to meet our users' need. 
+The below is a summary of what we currently support and links to individual guides.
+
+Check our [Public Roadmap](https://github.com/orgs/questdb/projects/1/views/5) to see tools we plan to support, and contribute to the items.
+
+Supported tools:
+
+- [Cube.js](https://questdb.io/blog/2022/04/26/time-series-data-analytics-with-questdb-and-cube/): A headless business intelligence platform
+- [Grafana](https://questdb.io/docs/third-party-tools/grafana/): An observability and monitoring application
+- [Kafka Connect](https://questdb.io/docs/third-party-tools/kafka): A event streaming tool
```
pswu11, line 16: What I understand is that QuestDB works with Kafka via Kafka Connect - is it more straightforward to actually just put Kafka? Maybe @javier has some insights.

docs/third-party-tools/overview.md
```diff
@@ -0,0 +1,20 @@
...
+
+We have a growing number of supported tools to meet our users' need. 
+The below is a summary of what we currently support and links to individual guides.
+
+Check our [Public Roadmap](https://github.com/orgs/questdb/projects/1/views/5) to see tools we plan to support, and contribute to the items.
+
+Supported tools:
+
+- [Cube.js](https://questdb.io/blog/2022/04/26/time-series-data-analytics-with-questdb-and-cube/): A headless business intelligence platform
+- [Grafana](https://questdb.io/docs/third-party-tools/grafana/): An observability and monitoring application
```
pswu11, line 15: Maybe to have some keywords to emphasize the data visualization (since that's mostly what users use Grafana for in this context)? Reference: https://grafana.com/docs/grafana/latest/introduction/

docs/third-party-tools/overview.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Third-party tools overview
+sidebar_label: Overview
+description: Introducing third-party tools that QuestDB supports
+---
+
+We have a growing number of supported tools to meet our users' need. 
+The below is a summary of what we currently support and links to individual guides.
+
+Check our [Public Roadmap](https://github.com/orgs/questdb/projects/1/views/5) to see tools we plan to support, and contribute to the items.
```
pswu11, line 10: > and contribute to the items.
this feels a bit vague, can we be a bit clear about we look forward to users' feedback/opinions and integration ideas?

pswu11: Also perhaps, we can put this below the list? 
the flow could be like:  
> try to find out what 3rd party tools work
> it's not on the list? oh here is some information about what might happen... and you can tell us what you expect!

amyshwang: Based on our previous discussion, I would park this idea for now because it might not be a good idea to list out tools that partially work, without specific instructions. I think it'd be a future task.

amyshwang:
Close as this is lower priority for now.

PR #1019, opened 2022-08-26
chore(docs) adding commit lag wording to `SELECT`

amyshwang:
- Tidying up existing info box as there are too many of them
- Adding commit lag wording to the tip box


docs/reference/sql/select.md
```diff
@@ -13,7 +13,9 @@ and evaluated from a table.
 
 :::tip
 
-The `TABLE` can either be a in your database (in which case you would pass the
+- While ingesting data using ILP, data may not be immediately available to a `SELECT` query, due to the [commit lag](/docs/guides/out-of-order-commit-lag) setting. Please see the [InfluxDB line protocol reference](/docs/reference/api/ilp/tcp-receiver#commit-strategy) for more information.
+
+- The `table` can either be a in your database (in which case you would pass the
```
pswu11, line 18: I'm aware that this is what it used to be but something seems a bit off here - maybe a word is missing?

amyshwang: I changed `TABLE` to `table` as the latter is used in the diagram.

jerrinot (approved):
looking good!

Issue #999, opened 2022-08-16
Update - outdated PG Wire Python example in  develop > insert data


amyshwang:
**Location:**
In [Insert data](https://questdb.io/docs/develop/insert-data/#postgresql-wire-protocol), the Python code block mentions:

```
This example uses the [psycopg2](https://github.com/psycopg/psycopg2) database adapter, which does not support prepared statements (bind variables). This functionality is on the roadmap for the antecedent [psychopg3](https://github.com/psycopg/psycopg3/projects/1) adapter.
```

The information is outdated and example needs to be updated

**Screenshots:**
![image](https://user-images.githubusercontent.com/107401731/184915733-f5c7826e-2119-4aac-891d-6f58d8625535.png)

amyshwang:
Using [psycopg documentation](https://www.psycopg.org/psycopg3/docs/basic/usage.html) I have this:


```
import psycopg as pg
import datetime as dt

# Connect to an existing QuestDB instance

with pg.connect("user=admin password=quest host=127.0.0.1 port=8812 dbname=qdb") as connection:
    # Open a cursor to perform database operations
    with connection.cursor() as cur:
        # Execute a command: this creates a new table
        cur.execute('''CREATE TABLE IF NOT EXISTS test_pg (
        ts TIMESTAMP, date DATE, name STRING, value INT)
        timestamp(ts);''')
        # Insert data into the table.

        # for x in range(10):
        #     now = dt.datetime.utcnow()
        #     date = dt.datetime.now().date()
        #     cur.execute("""
        #         INSERT INTO test_pg
        #             VALUES (%s, %s, %s, %s);
        #         """,
        #         (now, date, 'python example', x))
        cur.execute("INSERT INTO test_pg VALUES (2022, 2022, 'test test', 0);")
        # Query the database and obtain data as Python objects.
        cur.execute("SELECT * FROM test_pg")
        cur.fetchall()

        for record in cur:
            print(record)
        
    connection.commit()
```

Problems:
- I cannot get `for x in range` works
- My alternative `cur.execute("INSERT INTO test_pg VALUES (2022, 2022, 'test test', 0);")` works
- I cannot get python to print the result, so something here is not working:

```
        cur.execute("SELECT * FROM test_pg")
        cur.fetchall()

        for record in cur:
            print(record)
```
However the row can be found in the table when I access it from web console.

amyshwang:
I got this inserted:
[...]

PR #1025, opened 2022-09-02
Adding references to supported postgresql libraries

javier:
Added docs about psycopg2, pg_connect, and ruby-pg libraries.

Also reworded more explicitly the link to Postgresql compatibility docs


docs/reference/api/postgres.md
```diff
@@ -83,6 +83,13 @@ Support for `SELECT`, `INSERT`, `UPDATE`, `CREATE`, `DROP`, `TRUNCATE`.
 
 #### [pq](https://www.postgresql.org/docs/12/libpq.html) (C) `12`
 
+#### [Psycopg](https://www.psycopg.org/docs/) (Python) `2.9.3`
```
amyshwang, line 86: Based on my discussion with @amunra a while ago ([#999](https://github.com/questdb/questdb.io/issues/999), I think we now support psychopg3?

amyshwang (commented):
One question!

docs/reference/api/postgres.md
javier: I wasn't even aware psycopg3 was already released XD. Just tried it out and we seem to support it. Our Python exaples still show psycopg2, so maybe an opportunity to refresh the examples

javier: Just added the reference to psycopg3 on that page, but kept also the reference to psycopg2 as their official project page is still showing both

amyshwang: Yes #999 captures exactly that, the need to update the examples....

amyshwang (approved):

PR #1030, opened 2022-09-06
feat(docs) adding new blog post - forecasting and prediction

amyshwang:


amyshwang (commented):
As @javier mentioned, we could mention DBEngines here - “QuestDB is one of the most popular time series databases according to the independent reviewer DBEngines"

PR #1035, opened 2022-09-09
docs: remove sqlalchemy dependency of statuspage

gabor-boros:
## Description

The status page tutorial depends on SQLAlchemy which introduced a functionality that is Postgres specific. This patch removes the SQLAlchemy dependency.


amyshwang (commented):
L153 Voilá should be Voilà.

amyshwang (approved):

PR #1043, opened 2022-09-12
add trailing slash to navbars

argshook:
- add trailing slash to links in navbar and footer
- ensure trailing slash in docs sidebar


insmac (approved):
Looks and works good

PR #1051, opened 2022-09-20
refactor code of index page

argshook:
This PR adjusts the code of the index page in order for it to be easier
to maintain.

No visual or functional changes were done.

This is done for easier code reuse. For example, `<Section>` and
`<Section.Title>` is used instead of a something like `<div className={seCss["section__title--wide"]}>`

---

commit messages:
- refactor 'why questdb' section
- reuse Section for "Interactive Console" part
- extract QueryScroller to component
- use size="small" for title
- use Section component
- refactor feature tabs
- refactor use cases section


argshook:
Hmm github shows red status checks, even though things are green on netlify. Not really sure why that's the case

src/components/QueryScroller/index.tsx
```diff
@@ -0,0 +1,230 @@
+import React, { useCallback, useEffect, useState } from "react"
```
argshook, line 1: this file was extracted from `src/pages/index.tsx`

argshook:
closing in favor of #1052

PR #1052, opened 2022-09-21
refactor code of index page

argshook:
This PR adjusts the code of the index page in order for it to be easier
to maintain.

No visual or functional changes were done.

This is done for easier code reuse. For example, `<Section>` and
`<Section.Title>` is used instead of a something like `<div className={seCss["section__title--wide"]}>`

---

commit messages:
- refactor 'why questdb' section
- reuse Section for "Interactive Console" part
- extract QueryScroller to component
- use size="small" for title
- use Section component
- refactor feature tabs
- refactor use cases section


src/components/QueryScroller/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+    if (windowWidth != null && windowWidth < 800) {
+      setTop(M)
+      return
+    }
+
+    setTop(L)
+  }, [windowWidth])
+
+  return (
+    <section
```
insmac, line 85: Is it possible to use the `<Section>` component here?

src/components/QueryScroller/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+      className={clsx(
+        seCss.section,
+        seCss["section--inner"],
+        seCss["section--center"],
+        seCss["section--showcase"],
+      )}
+    >
+      <h2
```
insmac, line 93: Same for `<Section.Title>` ?

insmac (approved):
Minor comments

src/components/QueryScroller/index.tsx
argshook: i started to do tweaks here in this file and saw that it became quite big
so just wanted to leave this for a separate PR.

PR #1053, opened 2022-09-21
update text functions page with anchor for operators

argshook:
I noticed that the [`Text functions` documentation page](https://questdb.io/docs/reference/function/text) had links for each section except for the `~` regex matching operator.

This is because the heading is only one character tilde `~`, which is
special character that automatically gets removed from section links.

In addition, the `~` and `!~` could be regarded not as functions but as binary operators, due to a different syntax usage (for example `concat(str, ...)` vs `string ~ regex`).

Therefore, this PR puts `~` and `!~` under new section `Operators`. This
gives additional benefit of being able to link to this section.


argshook:
preview of the change: https://deploy-preview-1053--unruffled-blackwell-31bfb2.netlify.app/docs/reference/function/text/#operators

pswu11:
I'm not the best docs person to judge about the extra hierarchy - but since we also have an "operators" sector on the sidebar, would it be a bit confusing? 
I personally find it also neat to simply put all of the operators, functions, regex operators, ... on the same level. 
https://docs.databricks.com/sql/language-manual/sql-ref-functions-builtin-alpha.html


While other headings simply use the function itself - I'd tend to start with the actual operator just like other functions and include its meaning in the brackets. Would this allow the anchor? 

## `~` (match)


## `!~` (does not match)

argshook:
> I'm not the best docs person to judge about the extra hierarchy - but since we also have an "operators" sector on the sidebar, would it be a bit confusing? I personally find it also neat to simply put all of the operators, functions, regex operators, ... on the same level. https://docs.databricks.com/sql/language-manual/sql-ref-functions-builtin-alpha.html
> 
> While other headings simply use the function itself - I'd tend to start with the actual operator just like other functions and include its meaning in the brackets. WDYT? Would this allow the anchor link?
> 
> ## `~` (match)
> ## `!~` (does not match)

Yep, that would work. I updated the PR with this suggestion, thanks!

pswu11:
Sorry for being picky 😭  ... for readability, would it look clearer if we remove the code format on the title?
Just realized that the code format looks much smaller in headings 

Screenshot 2022-09-23 at 18 09 46

Screenshot 2022-09-23 at 18 09 51

WDYT?

argshook:
@pswu11 sure, no problem. Removed the code blocks

pswu11 (approved):
LGTM

PR #1059, opened 2022-09-23
add logos and fix their alignment

argshook:
In this PR:

- fix logo sizes in header marquee
- add Redpanda logo under `use with tools you love`
- fix Yahoo logo in Yahoo case study page


insmac (approved):
The only thing I'm not sure of is the Redpanda logo... maybe it would look better desaturated as per other ones?

CleanShot 2022-09-26 at 15 18 49@2x

PR #1061, opened 2022-09-26
typing effect index header

argshook:
This PR updates the header of the index page.

imageThe main update is the change of static image to a query typing animation. In addition, two new buttons: `Run this Query` and `Try another`:

- clicking `Run this Query` will open `demo.questdb.io` with editor pre-filled with shown query. It will also immediatelly run that query, so user will be able to see result in the grid (and the speed of execution)

- clicking `Try another` will show query animation again, but for another query.


insmac (approved):
Looks and works as intended 👍

src/modules/index-header/example-queries.ts
```diff
@@ -0,0 +1,146 @@
...
+  },
+
+  {
+    comment: [
+      "Find the latest received position for each ship in the ocean, for ships within a specific geo boundary and",
+      "within a time frame. The geo boundary is represented using a geohash of precission 3, which represents an area",
+      "of 156x156 Km. QuestDB supports geohashes from 5000x5000 km to 37.2mmx18.6mm. This dataset has been synthetically",
+      "generated for the purpose of this demo and contains over 122 million records",
+    ],
+    query: `SELECT * FROM pos LATEST BY id
```
puzpuzpuz, line 58: We should use LATEST ON syntax as LATEST BY is deprecated.

src/modules/index-header/example-queries.ts
```diff
@@ -0,0 +1,146 @@
...
+  {
+    comment: [
+      "Find the latest value available for each symbol. This dataset uses live crypto market data from Coinbase",
+    ],
+    query: `SELECT
+  *
+FROM
+  trades
+WHERE
+  symbol IN ('BTC-USD', 'ETH-USD') LATEST ON timestamp PARTITION BY symbol;`,
```
puzpuzpuz, line 11: nit: can we start LATEST ON/SAMPLE BY from a new line in all queries? These clauses are important enough to start with a new line for improved readability.

src/modules/index-header/example-queries.ts
```diff
@@ -0,0 +1,130 @@
+export default [
+  {
+    comment: [
+      "Find the latest value available for each symbol. This dataset uses live crypto market data from Coinbase",
+    ],
+    query: `SELECT * FROM trades
```
puzpuzpuz, line 6:
```suggestion
    query: `SELECT *
    FROM trades
```

src/modules/index-header/example-queries.ts
```diff
@@ -0,0 +1,130 @@
...
+  },
+
+  {
+    comment: [
+      "Find the latest received position for each ship in the ocean, for ships within a specific geo boundary and",
+      "within a time frame. The geo boundary is represented using a geohash of precission 3, which represents an area",
+      "of 156x156 Km. QuestDB supports geohashes from 5000x5000 km to 37.2mmx18.6mm. This dataset has been synthetically",
+      "generated for the purpose of this demo and contains over 122 million records",
+    ],
+    query: `SELECT * FROM pos LATEST ON id
```
puzpuzpuz, line 53: This one still uses LATEST ON.

src/modules/index-header/example-queries.ts
```diff
@@ -0,0 +1,130 @@
...
+  AND pickup_datetime <= '2015';`,
+  },
+
+  {
+    comment: [
+      "Count how many taxi rides per hour we saw in the 7 days from June 1st 2018. Note how we are using a SQL extension",
+      "to make working with time ranges more convenient. We are also using SAMPLE BY to aggregate records at regular intervals.",
+      "This dataset is open data from the City of New York and has over 1.6 billion records",
+    ],
+    query: `SELECT pickup_datetime, count() FROM trips
```
puzpuzpuz, line 77: nit:
```suggestion
    query: `SELECT pickup_datetime, count()
    FROM trips
```

src/modules/index-header/example-queries.ts
```diff
@@ -0,0 +1,130 @@
...
+  AND time < '2021-09-19T00:00:00.000000Z';`,
+  },
+
+  {
+    comment: [
+      "Get all the taxi rides during 2014 and 2015. Note how we are just using the year when filtering the time range.",
+      "QuestDB has SQL extensions for making working with timestamps easier, and we could filter by an arbitrary range",
+      "This dataset is open data from the City of New York and has over 1.6 billion records",
+    ],
+    query: `SELECT * FROM trips
```
puzpuzpuz, line 65: nit:
```suggestion
    query: `SELECT *
    FROM trips
```

src/modules/index-header/example-queries.ts
argshook: and that's correct, right? Because `LATEST BY` is deprecated?

src/modules/index-header/example-queries.ts
```diff
@@ -0,0 +1,133 @@
...
+  },
+
+  {
+    comment: [
+      "Find the latest received position for each ship in the ocean, for ships within a specific geo boundary and",
+      "within a time frame. The geo boundary is represented using a geohash of precission 3, which represents an area",
+      "of 156x156 Km. QuestDB supports geohashes from 5000x5000 km to 37.2mmx18.6mm. This dataset has been synthetically",
+      "generated for the purpose of this demo and contains over 122 million records",
+    ],
+    query: `SELECT * FROM pos LATEST ON id
```
puzpuzpuz, line 54: LATEST ON has to follow WHERE clause:
```sql
SELECT *
FROM pos
WHERE geo6 within(#wtq)
  AND time < '2021-09-19T00:00:00.000000Z'
LATEST ON time PARTITION BY id;
```

src/modules/index-header/example-queries.ts
```diff
@@ -0,0 +1,133 @@
...
+  AND time < '2021-09-19T00:00:00.000000Z';`,
+  },
+
+  {
+    comment: [
+      "Get all the taxi rides during 2014 and 2015. Note how we are just using the year when filtering the time range.",
+      "QuestDB has SQL extensions for making working with timestamps easier, and we could filter by an arbitrary range",
+      "This dataset is open data from the City of New York and has over 1.6 billion records",
+    ],
+    query: `SELECT *
```
puzpuzpuz, line 67: This query doesn't do what it has to do according to the description - the maximum pickup date specified by the filter is `2015-01-01`, instead of end of 2015. The correct filter has to be something like `pickup_datetime BETWEEN '2014-01-01T00:00:00.000000Z' AND '2015-12-31T23:59:59.999999Z'` or `pickup_datetime >= '2014-01-01' AND pickup_datetime < '2016-01-01'`.

src/modules/index-header/example-queries.ts
puzpuzpuz: My bad, I meant that the syntax was incorrect - LATEST ON must follow WHERE. It's fixed now.

puzpuzpuz (approved):

PR #1063, opened 2022-09-27
chore: reducing ILP writer idle timeout defaults

glasstiger:
Doc changes for PR: https://github.com/questdb/questdb/pull/2548

-------------------------------------------------------------------------
Reducing ILP writer idle timeout default settings so users would not have to wait 20-30 seconds for the ingested data to be available for queries.
With the new defaults the data becomes visible within 1.5 seconds after ingestion stops.
```
line.tcp.maintenance.job.interval:          30000 -> 1000
line.tcp.min.idle.ms.before.writer.release: 10000 ->  500
```


puzpuzpuz (approved):

docs/reference/api/ilp/tcp-receiver.md
```diff
@@ -135,7 +135,7 @@ interval can be set by the below configuration parameter.
 line.tcp.commit.interval.fraction=0.2
 ```
 
-If the fraction value is set to `0`, the
+If the fraction value and/or commit lag is set to `0`, the
```
amyshwang, line 138: Would this be more concise?

```suggestion
If the result of commit interval is `0`, the
```

docs/reference/configuration.md
```diff
@@ -469,8 +469,8 @@ line protocol.
...
-| line.tcp.maintenance.job.interval          | 30000        | Maximum amount of time (in milliseconds) between maintenance jobs, these will commit any uncommitted data on inactive tables.                                                                                                                         |
-| line.tcp.min.idle.ms.before.writer.release | 30000        | Minimum amount of idle time before a table writer is released in milliseconds.                                                                                                                                                                        |
+| line.tcp.maintenance.job.interval          | 1000         | Maximum amount of time (in milliseconds) between maintenance jobs, these will commit any uncommitted data on inactive tables.                                                                                                                         |
```
amyshwang, line 472:
```suggestion
| line.tcp.maintenance.job.interval          | 1000         | Maximum amount of time (in milliseconds) between maintenance jobs committing any uncommitted data on inactive tables.                                                                                                                         |
```

docs/reference/configuration.md
```diff
@@ -469,8 +469,8 @@ line protocol.
...
+| line.tcp.min.idle.ms.before.writer.release | 500          | Minimum amount of idle time before a table writer is released in milliseconds.                                                                                                                                                                        |
```
amyshwang, line 473:
```suggestion
| line.tcp.min.idle.ms.before.writer.release | 500          | Minimum amount of idle time (in milliseconds) before a table writer is released.                                                                                                                                                                        |
```

docs/develop/insert-data.md
```diff
@@ -330,7 +330,6 @@ If you don't see your inserted data, this is usually down to one of two things:
   settings to commit data more frequently.
   ```ini title=server.conf
   cairo.max.uncommitted.rows=1
-  line.tcp.maintenance.job.interval=100
```
amyshwang, line 333: My understanding is that we're showing an example of high put-through. Is it correct to delete the setting here? (As this change is irrelevant to the default idle time?)

amyshwang (changes requested):
Made some suggestions and add one question.

docs/develop/insert-data.md
glasstiger: ILP commits every single row as they are ingested when `cairo.max.uncommitted.rows` is set to 1.
In this case writer idle timeout and maintenance job interval settings are irrelevant, no need to change them.
Other option would be to leave `cairo.max.uncommitted.rows` as it is and lower writer idle timeout and maintenance job interval settings instead.
To be honest, i am not even sure we need this in the doc at all with the new writer idle timeout settings.
Maybe we could just remove it completely?
User will see the data in 1.5 seconds after ILP ingestion stops.

amyshwang: So with `cairo.max.uncommitted.rows` set to 1, every single row will be committed except for the latest row - if (albeit unlikely) there is no need data coming in, it'd stay uncommitted until the conditions of other configurations are met. I guess this is why the interval is mentioned here?!
But I agree that based on the reduced idle time default, this can go:

> For development (and development only), you may want to tweak configuration
>   settings to commit data more frequently.
>   ```ini title=server.conf
>   cairo.max.uncommitted.rows=1
>   line.tcp.maintenance.job.interval=100
> ```

amyshwang: Thanks. Happy to maintain your change here. I added [656f0b4](https://github.com/questdb/questdb.io/pull/1063/commits/656f0b43f47120bd02aba9998de9faac4de75ba3) to tweak the `cairo.max.uncommitted.rows` wording.

amyshwang (approved):

PR #1070, opened 2022-09-28
chore(docs): adding FAQ for changing column data type

amyshwang:


docs/troubleshooting/faq.md
```diff
@@ -11,14 +11,40 @@ issues met when running QuestDB, as well as solutions to them.
...
+   `query.timeout.sec`.
+1. Use `UPDATE` to copy the existing column content to the new column. Now, the
+   column has the correct content with the new data type.
+1. Delete the old column.
+1. Rename the new column accordingly.
+
+For example, to change `old_col` from `STRING` to `SYMBOL` for table `x`:
+
+```questdb-sql
+
```
puzpuzpuz, line 40: nit: redundant newline here and below.

docs/troubleshooting/faq.md
```diff
@@ -11,14 +11,40 @@ issues met when running QuestDB, as well as solutions to them.
...
+ALTER TABLE x ADD COLUMN new_col SYMBOL;
```
puzpuzpuz, line 41: nit: shall we use `my_table` instead of `x` as the table name?

docs/troubleshooting/faq.md
```diff
@@ -11,14 +11,40 @@ issues met when running QuestDB, as well as solutions to them.
...
 ## How do I delete a row?
 
 See our guide on [modifying data](/docs/guides/modifying-data).
 
+## How do I change the data type of a column?
+
+The SQL `UPDATE` keyword can be used to change the data type of a column.
+
+1. Add a new column to the table and define the desired data type.
+1. Stop data ingestion and increase
```
puzpuzpuz, line 29: We should suggest some value for `query.timeout.sec` and also explain why the user should change it. Also, it might be a good idea to remind the user that the database has to be restarted to put the changed setting in force.

docs/troubleshooting/faq.md
```diff
@@ -11,14 +11,40 @@ issues met when running QuestDB, as well as solutions to them.
...
+[guide to commit lag](/docs/guides/out-of-order-commit-lag) to understand the
+concept and
 [InfluxDB line protocol reference](/docs/reference/api/ilp/tcp-receiver#commit-strategy)
 to understand these rules.
 
 ## How do I delete a row?
 
 See our guide on [modifying data](/docs/guides/modifying-data).
 
+## How do I change the data type of a column?
```
puzpuzpuz, line 24: nit: I'd narrow down the question to a particular use case to help users understand it better:
```suggestion
## How do I convert a STRING column to a SYMBOL or vice versa?
```

docs/troubleshooting/faq.md
```diff
@@ -11,14 +11,40 @@ issues met when running QuestDB, as well as solutions to them.
...
+
+The SQL `UPDATE` keyword can be used to change the data type of a column.
```
puzpuzpuz, line 26: nit: we should also mention that the same approach may be used to increase capacity of a SYMBOL column if it was previously undersized.

docs/troubleshooting/faq.md
amyshwang: Are we only dealing with these two type? How about integer and other types?

puzpuzpuz: No, but this use case is the most frequent one. If we keep the title as is, it would be nice to describe this use case below, as well as changing the symbol capacity one.

docs/troubleshooting/faq.md
amyshwang: The default is 60 secs and I guess depending on the size of the column this should be increased? What would be a reasonable setting? 
Also do we recommend them to change the setting back after the restart?

puzpuzpuz: Yes, we should recommend setting it to a very high value, say, an hour (UPDATE may take a while to get done) and revert it afterwards. The default value serves as a protection from running super heavy (and super slow) queries on the server and, hence, leading to starvation of pretty much every other tasks.

docs/troubleshooting/faq.md
```diff
@@ -11,14 +11,49 @@ issues met when running QuestDB, as well as solutions to them.
...
+   `query.timeout.sec`, as `UPDATE` may take a while to complete. Depending on
+   the size of the column, we recommend to increase the value significantly: the
+   default is 60 seconds and it may be reasonable to increase it to one hour.
+   Restart the instance after changing the configuration, to activate the
+   change.
+1. Use `UPDATE` to copy the existing column content to the new column. Now, the
+   column has the correct content with the new data type.
+1. Delete the old column.
+1. Rename the new column accordingly.
+
```
puzpuzpuz, line 45:
```suggestion
1. Set the `query.timeout.sec` setting to the original value and start data ingestion.

```

docs/troubleshooting/faq.md
```diff
@@ -11,14 +11,49 @@ issues met when running QuestDB, as well as solutions to them.
...
+For example, to change `old_col` from `STRING` to `SYMBOL` for table `my_table`:
+
+```questdb-sql
+ALTER TABLE my_table ADD COLUMN new_col SYMBOL;
+-- increase `query.timeout.sec` and restart QuestDB.
```
puzpuzpuz, line 50: nit: I'd remove the comments since we have the same content in the above instruction. WDYT?

docs/troubleshooting/faq.md
```diff
@@ -11,14 +11,49 @@ issues met when running QuestDB, as well as solutions to them.
...
 
 ## How do I delete a row?
 
 See our guide on [modifying data](/docs/guides/modifying-data).
 
+## How do I convert a `STRING` column to a `SYMBOL` or vice versa?
+
+The SQL `UPDATE` keyword can be used to change the data type of a column. The
+same approach can also be used to increase the
+[capacity of a `SYMBOL` column](/docs/concept/symbol/#symbol-columns/) that is
```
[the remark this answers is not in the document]
amyshwang: It works in the preview and trailing slash helps with SEO. I will remove the first slash perhaps.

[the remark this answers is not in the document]
puzpuzpuz: No, the anchor doesn't work in the preview: it should navigate to the `Usage of symbols` section, but it doesn't. That's because the anchor shouldn't end with a slash, i.e. it should be `#symbol-columns` instead of `#symbol-columns/`.

PR #1073, opened 2022-09-29
inject SVG into DOM

argshook:
This PR includes a number of small performance tweaks:

- update browserslist
- set fixed size for integration icons
- refactor `<Customers />` to inject SVG instead of loading through HTTP
- remove unsupported namespace tags
- inject logos into DOM
- add viewbox to SVGs that were missing it
- remove unused CSS

No intentional visual changes or functional changes in this PR


insmac (approved):
Looking and working ood, the only thing I remember is that don't we get performance / page index degradation by introducing more DOM elements with inline SVGs? (not sure, just asking)

argshook:
> Looking and working OK, the only thing I remember is that don't we get performance / page index degradation by introducing more DOM elements with embedded SVGs? (not sure, just asking)

yeah that's true and i actually moved those icons to come from files in earlier PRs via `<img src="...`. But then i noticed that on slower network and especially on mobile, each svg takes time to load and website feels slower, you can see icons loading one by one.

So it's a tradeoff, we do get larger DOM but much less HTTP requests.

This is current questdb.io network requests filtered by "svg":

image

and this is the same thing except recording network requests from the preview link of this PR:

image

---

Adding SVGs inline is not ideal, but we have bigger problems to solve before "SVG bloating the DOM" becomes an issue.

For example:

image

PR #1098, opened 2022-10-04
add `open live demo` button to index page header code preview

argshook:


argshook:
Closing, we discussed that this is not necessary

Issue #1078, opened 2022-09-30
Update - Capturing `like/llike` functions


amyshwang:
**Describe the requested changes:**

In the core repo QuestDB has two regex functions `like` and `ilike` that are not yet documented:
https://github.com/questdb/questdb/tree/master/core/src/main/java/io/questdb/griffin/engine/functions/regex

io.questdb.griffin.engine.functions.regex.LikeStrFunctionFactory
io.questdb.griffin.engine.functions.regex.ILikeStrFunctionFactory
io.questdb.griffin.engine.functions.regex.LikeCharFunctionFactory

The task involves:
Create a new section for each function, following the [template](https://github.com/questdb/questdb.io/blob/master/docs/__guidelines/template/function.md).

**Location:**
/docs/reference/operators/regex/

orensaldanha:
Can I take this?

PR #1103, opened 2022-10-05
Capturing like/llike functions in /docs/reference/operations/regex.md #1078

orensaldanha:
#1078 

I have added documentation for the LIKE operator.

However, the ILIKE operator is not working as expected.

Screenshot from 2022-10-04 21-44-40


puzpuzpuz:
> However, the ILIKE operator is not working as expected.

Our ILIKE implementation does a case-sensitive matching. This aspect should be documented.

orensaldanha:
So the LIKE and ILIKE operator have the same functionality?
1) Should I copy the LIKE documentation for ILIKE as it is 
2) or should I make a small section for ILIKE operator that says ILIKE makes a case sensitive match similiar to LIKE , adding that  this is contrary to other database systems like Postgres which make a case insensitive match for ILIKE

puzpuzpuz:
As far as I can see in the code, LIKE and ILIKE use the same implementation, so ATM their behavior is the same.

puzpuzpuz:
I've created https://github.com/questdb/questdb/issues/2585 to track the ILIKE issue. Yet, I think we should document it properly, i.e. as case-insensitive. The bug should be fixed soon.

docs/reference/operators/regex.md
```diff
@@ -20,3 +20,36 @@ matching via regular expressions.
 `string !~ regex` - checks if `string` value does not match regex
 
 `symbol !~ regex` - checks if `symbol` value does not match regex
+
+## LIKE 
+
+The `LIKE` operator returns true if it matches the specified pattern
```
amyshwang, line 26: 1. Please follow [the template](https://github.com/questdb/questdb.io/blob/master/docs/__guidelines/template/function.md#function_name) to document this:
 
'LIKE' - returns...
 This applies to the rest of the PR.
2. I think it is not just returns true but returns the row?

amyshwang (changes requested):
Thank you for your contribution and the discussion on the bug! 😃 
Could you document the desired behaviours for both `LIKE` and 'ILIKE` (probably in the same section to make it concise but listing out the difference in a clear manner with examples). We can then mark this task completed, independent of whether the bug is fixed.

docs/reference/operators/regex.md
orensaldanha: The operator itself returns a boolean but the WHERE clause returns the row if the operator result is true

Screenshot from 2022-10-05 15-55-46

docs/reference/operators/regex.md
```diff
@@ -20,3 +20,71 @@ matching via regular expressions.
...
+`symbol` is [string].
+
+
+### Return value
+
+Return value type is [boolean].
+
+### Description
+
+The pattern can contain wildcards which are interpreted as follows:
```
amyshwang, line 41:
```suggestion
The Java regex syntax can contain wildcards which are interpreted as follows:
```

docs/reference/operators/regex.md
```diff
@@ -20,3 +20,71 @@ matching via regular expressions.
...
+- `?` - matches any single character
+- `%` - matches any sequence of zero or more characters
+
+Wildcards can be used as follows:
+
+|      query             | result  |
+| ---------------------- | ------- |
+| 'quest' LIKE 'quest'   |  true   |
```
amyshwang, line 49: Applying this to all:
```suggestion
|  SELECT 'quest' LIKE 'quest'   |  true   |
```

In this way, users can copy it and run it on the demo site directly.

amyshwang (changes requested):
Thank you for the update - more feedback on this!

docs/reference/operators/regex.md
```diff
@@ -20,3 +20,71 @@ matching via regular expressions.
...
+
+## LIKE / ILIKE
+
+`string LIKE pattern` - returns true if `string` value matches `pattern` otherwise returns false (case sensitive match)
+
+`string ILIKE pattern` - returns true if `string` value matches `pattern` otherwise returns false (case-insensitive match)
+
+### Arguments
+`string` is [string].
+`symbol` is [string].
```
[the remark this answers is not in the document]
orensaldanha: @amyshwang correct me if I'm wrong, but I don't think "pattern following the Java regex syntax" is an appropriate description for pattern here since the pattern does not support the whole range of regex meta characters rather it supports only the two wildcards '?' and '%'.

docs/reference/operators/regex.md
```diff
@@ -20,3 +20,74 @@ matching via regular expressions.
...
+
+## LIKE / ILIKE
+
+`(string) LIKE (pattern)` - returns true if the `string` value matches `pattern`, otherwise returns false (case sensitive match).
+
+`(string) ILIKE (pattern)` - returns true if the `string` value matches `pattern`, otherwise returns false (case-insensitive match).
+
+### Arguments
+
+`string` is a string which is commonly the name of a column.
```
amyshwang, line 32:
```suggestion
`string` is an expression that evaluates to the `string` data type. 
```

docs/reference/operators/regex.md
```diff
@@ -20,3 +20,74 @@ matching via regular expressions.
...
+
+Return value type is `boolean`.
+
+### Description
+
+If the pattern doesn't contain wildcards, then the pattern represents the string itself.
+
+The wildcards whuch can be used in pattern are interpreted as follows:
+- `?` - matches any single character
+- `%` - matches any sequence of zero or more characters
```
amyshwang, line 46: Missing full stops.

docs/reference/operators/regex.md
```diff
@@ -20,3 +20,74 @@ matching via regular expressions.
...
+`ILIKE` performs a case-insensitive match as follows:
+
+|            query              | result  |
+| ----------------------------- | ------- |
+| SELECT 'quest' ILIKE 'QUEST'  |  true   |
+| SELECT 'qUeSt' ILIKE 'QUEST'  |  true   |
+| SELECT 'quest' ILIKE 'QUE%'   |  true   |
+| SELECT 'QUEST' ILIKE '\_ues_' |  true   |
+
+### Example
```
amyshwang, line 67:
```suggestion
### Examples
```

docs/reference/operators/regex.md
```diff
@@ -20,3 +20,74 @@ matching via regular expressions.
...
+
+#### ILIKE
+
+```questdb-sql 
+SELECT * FROM trades
+WHERE symbol ILIKE '%-usd'
+LATEST ON timestamp PARTITION BY symbol;
+```
+
+| symbol | side | price | amount | timestamp |
```
amyshwang, line 90: As mentioned, we'd like proper behaviours to be documented so this is not correct. In addition, https://github.com/questdb/questdb/issues/2585 has been merged.

amyshwang (changes requested):
Nearly there 👍

docs/reference/operators/regex.md
orensaldanha: By proper behaviour do you mean:
1) the result it returns on the demo site (which is still not updated with the correct ILIKE implementation).
2) the assumed correct behaviour for the ILIKE operator.

In my PR, I have used the assumed behaviour.

amyshwang: You're right! My bad!

docs/reference/operators/regex.md
```diff
@@ -21,7 +21,76 @@ matching via regular expressions.
...
+`pattern` is a pattern which can contain wildcards like `?` and `%`.
+
+### Return value
+
+Return value type is `boolean`.
+
+### Description
+
+If the pattern doesn't contain wildcards, then the pattern represents the string itself.
+
```
bziobrowski, line 43: typo: `whuch`

docs/reference/operators/regex.md
```diff
@@ -21,7 +21,76 @@ matching via regular expressions.
...
+The wildcards whuch can be used in pattern are interpreted as follows:
+- `?` - matches any single character.
+- `%` - matches any sequence of zero or more characters.
+
+Wildcards can be used as follows:
+
+|            query              | result  |
+| ----------------------------- | ------- |
+| SELECT 'quest' LIKE 'quest'   |  true   |
```
bziobrowski, line 52: Does : 
```sql
SELECT 'quest' LIKE 'quest_'
```
really return true ? 

It doesn't seem to be the case on master branch or demo instance .

orensaldanha: Yep my bad, 

it's supposed to be SELECT 'quest' LIKE 'ques_'

orensaldanha: As for the second one, 
I intended it to be: SELECT 'quest' LIKE '_ues_'
I added the '\' to escape the underscore as it was displaying it in italics.

Screenshot from 2022-10-11 15-22-46

docs/reference/operators/regex.md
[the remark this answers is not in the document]
amyshwang: Yes you're right! Apologies for the confusion. But I think it should be "_" and "%"?

amyshwang (changes requested):

docs/reference/operators/regex.md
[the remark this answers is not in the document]
orensaldanha: Yep my bad. It's '_' and not '?'

docs/reference/operators/regex.md
```diff
@@ -21,7 +21,76 @@ matching via regular expressions.
...
 
+`(string) LIKE (pattern)` - returns true if the `string` value matches `pattern`, otherwise returns false (case sensitive match).
+
+`(string) ILIKE (pattern)` - returns true if the `string` value matches `pattern`, otherwise returns false (case-insensitive match).
+
+### Arguments
+
+`string` is an expression that evaluates to the `string` data type.
+
+`pattern` is a pattern which can contain wildcards like `?` and `%`.
```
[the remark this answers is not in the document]
orensaldanha: Yep made a mistake. Sorry for the confusion.

docs/reference/operators/regex.md
amyshwang: Could you please revert the change? We need to check the rendering, not the source!

orensaldanha: yep done

amyshwang (approved):
Thank you for your contribution!

PR #1106, opened 2022-10-05
chore(docs): suggest using SSD on import pages

puzpuzpuz:


docs/guides/importing-data.md
```diff
@@ -5,17 +5,24 @@ description:
...
+:::caution
 
-:::note
-
-This guide is applicable for QuestDB version 6.5 and higher.
+For partitioned tables, the best `COPY` performance can be achieved only on a
+machine with a local, physically attached SSD. If you're fine with 5-10x longer
+import times, you may use a network block storage, such as AWS EBS, but make
+sure to configure the maximum IOPS and throughput setting values for your
+volume.
```
amyshwang, line 14: Could we move this to line 25 so that we at least have some text before the warning?

```suggestion
machine with a local, physically attached SSD. It is possible to use a network block storage, such as AWS EBS to perform the operation, with the following impact:
- Users need to configure the maximum IOPS and throughput setting values for the volume.
- The required import time is likely to be 5-10x longer.
```

puzpuzpuz: Addressed in 8fc67f9d1e11365b74a7ad9fea0d32afe8ab4931

amyshwang (approved):

PR #1116, opened 2022-10-07
Fix documentation inaccuracy (close #1060)

### LINKED ISSUE #1060: Bug - Misleading partitioning info in ILP
The [ILP reference](https://questdb.io/docs/reference/api/ilp/overview/#behavior) says: 
> The timestamp column is automatically created as [designated timestamp](https://questdb.io/docs/concept/designated-timestamp) with the [partition strategy](https://questdb.io/docs/concept/partitions) set to NONE. If you would like to define a partition strategy, you should [CREATE](https://questdb.io/docs/reference/sql/create-table) the table beforehand.

I believe this is wrong. It'll use the default global partitioning which is set as by-day by default.

3t8:
Change documentation for default partition strategy
#1060


docs/reference/api/ilp/overview.md
```diff
@@ -54,8 +54,8 @@ Each ILP message has to end with new line `\n` character.
   automatically recognized and assigned based on the data.
 - The `timestamp` column is automatically created as
   [designated timestamp](/docs/concept/designated-timestamp) with the
-  [partition strategy](/docs/concept/partitions) set to `NONE`. If you would
-  like to define a partition strategy, you should
+  [partition strategy](/docs/concept/partitions) set to `DAY`. If you would
```
amyshwang, line 57: The partition strategy page has similar wording that needs changing - please ensure that consistency is maintained throughout the document.

amyshwang (changes requested):
This default is used in ingestion via ILP: 
https://questdb.io/docs/reference/configuration/#influxdb-line-protocol
image

Could we also incorporate the info into the paragraph?

amyshwang (changes requested):
Thank you for your submission! 
Please ensure consistency across the whole Docs and I added a couple of comments too.

docs/concept/partitions.md
```diff
@@ -21,7 +21,7 @@ import Screenshot from "@theme/Screenshot"
 ## Properties
 
 - Available partition intervals are `NONE`, `YEAR`, `MONTH`, `DAY`, and `HOUR`.
-- Default behavior is `PARTITION BY NONE`.
+- Default behavior is `PARTITION BY DAY`.
```
jerrinot, line 24: this is misleading. it's BY DAY for tables implicitly created via ILP ingestion and NONE for tables explicitly created via SQL.

docs/reference/sql/create-table.md
```diff
@@ -169,7 +169,7 @@ created.
 
 `PARTITION BY` allows for specifying the
 [partitioning strategy](/docs/concept/partitions) for the table. The default
-partitioning strategy is `NONE` and tables can be partitioned by one of the
+partitioning strategy is `DAY` and tables can be partitioned by one of the
```
jerrinot, line 172: I believe this change should not be done at all. This is the reference documentation for SQL. Tables created via SQL have default partitioning NONE.

amyshwang: How about changing it to "By default, tables do not have partition.

docs/concept/partitions.md
amyshwang: We can list both and specify the context.

amyshwang:
Hi this PR has not been active for a while, I will close this.

PR #1124, opened 2022-10-11
chore(docs) Tidying up regexp wording.

amyshwang:


docs/reference/operators/regex.md
```diff
@@ -5,37 +5,37 @@ description: Regex operators reference documentation.
 ---
 
 This page describes the available operators to assist with performing pattern
-matching via regular expressions.
+matching via regular expressions. QuestDB uses
```
bziobrowski, line 8: java.util.regex might not be clear for someone not familiar with Java so how about : 
[java.util.regex implementation] 
or 
[Java regular expression implementation] 
?

docs/reference/operators/pattern-matching.md
```diff
@@ -0,0 +1,127 @@
...
+- `(string) ~ (regex)` - returns true if the `string` value matches the `regex` rules, otherwise returns false (case sensitive match).
+- `(string) !~ (regex)` - returns true if the `string` value fails to match the `regex` rules, otherwise returns false (case sensitive match).
+
+### Arguments
+
+- `string` is an expression that evaluates to the `string` data type.
+- `regex` is any regular expression pattern used to calculate the rules.
+
+## LIKE/ILIKE
+
```
bziobrowski, line 22: case sensitive  vs "case-insensitive" below

docs/reference/operators/pattern-matching.md
```diff
@@ -0,0 +1,127 @@
...
+sidebar_label: Pattern matching
+description: Pattern matching operators reference documentation.
+---
+
+This page describes the available operators to assist with performing pattern
+matching. For operators using regular expressions (`regex` in the syntax), QuestDB uses
+[Java regular expression implementation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Pattern.html).
+
+## ~ (match) and !~ (does not match)
+
```
bziobrowski, line 12: case sensitive vs "case-insensitive" below

docs/reference/operators/pattern-matching.md
```diff
@@ -0,0 +1,127 @@
...
+- `(string) ~ (regex)` - returns true if the `string` value matches the `regex` rules, otherwise returns false (case sensitive match).
```
bziobrowski, line 13: I'd double if "matching regex rules" is a valid or popular expression . 
I think normally something simply matches a regular expression .

docs/reference/operators/pattern-matching.md
```diff
@@ -0,0 +1,127 @@
...
+-- Returns true
+
+SELECT 'quest' LIKE 'ques_';   
+-- Returns true   
+
+SELECT 'quest' LIKE 'que%';
+-- Returns true   
+
+SELECT 'quest' LIKE '_ues_';
+-- Returns true
```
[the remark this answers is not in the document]
amyshwang: Updated - removing empty lines.

docs/reference/operators/pattern-matching.md
```diff
@@ -0,0 +1,124 @@
...
+matching. For operators using regular expressions (`regex` in the syntax), QuestDB uses
+[Java regular expression implementation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Pattern.html).
+
+## ~ (match) and !~ (does not match)
+
+- `(string) ~ (regex)` - returns true if the `string` value matches a regular expression, `regex`, otherwise returns false (case sensitive match).
+- `(string) !~ (regex)` - returns true if the `string` value fails to match a regular expression, `regex`, otherwise returns false (case sensitive match).
+
+### Arguments
+
```
bziobrowski, line 17: I'm not sure I understand what 'calculate the rules' means in this context .

docs/reference/operators/pattern-matching.md
```diff
@@ -0,0 +1,124 @@
...
+## LIKE/ILIKE
+
+- `(string) LIKE (pattern)` - returns true if the `string` value matches `pattern`, otherwise returns false (case sensitive match).
+- `(string) ILIKE (pattern)` - returns true if the `string` value matches `pattern`, otherwise returns false (case insensitive match).
+
+### Arguments
+
+- `string` is an expression that evaluates to the `string` data type.
+- `pattern` is a pattern which can contain wildcards like `_` and `%`.
+
```
bziobrowski, line 30: Is this return value block shared by all operators mentioned above ?

bziobrowski: Might be worth to mention that there's a difference between ~/!~ and like/ilike in that :
- ~ !~ try to match any part of the input string  
- like/like try to match whole string 

Example:

```sql
select 'quest' ~ 'ue'  
--true

select 'quest' like 'ue'  
--false
```

bziobrowski (approved):

PR #1132, opened 2022-10-14
chore(docs): Update the documentation of ALTER DROP PARTITION 

marregui:
With PR https://github.com/questdb/questdb/pull/2606 the active partition can now be dropped, and thus I have updated the documentation to remove an info block which informed users that the active partition could not be dropped.

Active partition: most chronologically recent partition to which data are being appended to.


amyshwang (approved):
Thank you!

PR #1151, opened 2022-10-24
chore(docs) adding position as an alias 

amyshwang:
Captured #2599
Refactored existing content


docs/reference/function/text.md
```diff
@@ -128,10 +128,10 @@ SELECT name, right('Thompson', 2) r FROM names LIMIT 3
 | AMELIE | IE  |
 | TOM    | OM  |
 
-## strpos
+## strpos(position)
```
puzpuzpuz, line 131: nit: this reads as if `strpos` had a single argument called `position`. How about `strpos / position`?

puzpuzpuz (approved):

docs/reference/function/text.md
amyshwang:
```suggestion
## strpos / position
```

PR #1155, opened 2022-10-25
chore(docs) Adding `CREATE TABLE LIKE`

amyshwang:
https://github.com/questdb/questdb/pull/2580
Diagram:
image


docs/reference/sql/create-table.md
```diff
@@ -148,6 +148,48 @@ CREATE TABLE my_table(symb SYMBOL INDEX CAPACITY 128, price DOUBLE, ts TIMESTAMP
 See [Index](/docs/concept/indexes#how-indexes-work) for more information about
 index capacity.
 
+## CREATE TABLE LIKE
+
+The `LIKE` keyword clones the table schema of an existing table. Setting such as designated timestamp, commit lag, and index capacity will be cloned, too.
```
puzpuzpuz, line 153: nit:
```suggestion
The `LIKE` keyword clones the table schema of an existing table. Table settings and parameters such as designated timestamp, symbol column indexes, commit lag, and index capacity will be cloned, too. 
```

docs/reference/sql/create-table.md
```diff
@@ -311,27 +316,4 @@ CREATE TABLE taxi_trips AS(
   SELECT * FROM taxi_trips_unordered ORDER BY pickup_time
 ) TIMESTAMP(pickup_time)
 PARTITION BY MONTH;
-```
-
-### CREATE TABLE WITH parameters
```
puzpuzpuz, line 316: This section is still relevant, so we should keep it. The only improvement here should be to clearly state that `CREATE TABLE ... (LIKE ...)` doesn't support WITH or any other additional clause.

static/img/docs/diagrams/.railroad
```diff
@@ -1,11 +1,13 @@
...
   '(' columnName typeDef (',' columnName typeDef)* ')'
   ) | (
-  'as' '(' selectSql ')'  (',' castDef)* (',' indexDef)*
-  )
+  'AS' '(' selectSql ')'  (',' castDef)* (',' indexDef)*
+  )| (
+ '(' 'LIKE' sourceTableName ')'
+)
   ) (
   ('timestamp' '(' columnRef ')' ) ('PARTITION' 'BY' ('NONE' | 'YEAR' | 'MONTH' | 'DAY' | 'HOUR' ))?
```
puzpuzpuz, line 12: ATM `CREATE TABLE ... (LIKE ...)` doesn't support there additional clauses, so we need a separate syntax diagram for it.

docs/reference/sql/create-table.md
```diff
@@ -148,6 +148,48 @@ CREATE TABLE my_table(symb SYMBOL INDEX CAPACITY 128, price DOUBLE, ts TIMESTAMP
...
+    rnd_symbol('ABB', 'CDD') c
+  FROM
+    long_sequence(100)
+  WHERE false
+);
+```
+
+:::note
+
+Notice the `WHERE false` condition.
```
puzpuzpuz, line 177: Do we really need the above weird example considering that we have CREATE TABLE LIKE now?

docs/reference/sql/create-table.md
amyshwang: I deleted it because I think it is a repetition of L226 ## WITH table parameters (just before examples) and I think it was not adding a lot of new information. Do you think I should revert it?

docs/reference/sql/create-table.md
```diff
@@ -263,6 +256,16 @@ For more information on commit lag and the maximum uncommitted rows, see the
 guide for [out-of-order commits](/docs/guides/out-of-order-commit-lag) and
 [ILP commit strategy](/docs/reference/api/ilp/tcp-receiver#commit-strategy).
 
+## CREATE TABLE LIKE
+
+The `LIKE` keyword clones the table schema of an existing table. Table settings
```
puzpuzpuz, line 261: nit:
```suggestion
The `LIKE` keyword clones the table schema of an existing table without copying the data. Table settings
```

docs/reference/sql/create-table.md
puzpuzpuz: I see. No need to revert then.

puzpuzpuz (approved):

PR #1162, opened 2022-10-28
chore(docs) adding literal to data type

amyshwang:


amyshwang:
Need more thorough investigation - task captured in #1164

PR #1165, opened 2022-11-01
chore(docs) document Kafka ILP connector

jerrinot:


docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+### QuestDB Symbol Type
+
+QuestDB supports a special type called [Symbol](https://questdb.io/docs/concept/symbol/). Use the `symbols`
+configuration option to specify which columns should be created as the `symbol` type
+
+### Target Table Considerations
+
+When a target table does not exist in QuestDB then it will be automatically created when a first row arrives. This is
+recommended approach for development and testing.
+
```
amyshwang, line 217: Should "arrives" be changed to "is committed" to be more accurate? Is this correct at all?

```suggestion
When a target table does not exist in QuestDB, it will be automatically created when the first row arrives. This approach is only recommended for development and testing.
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+In production, it's recommended to [create tables manually via SQL](https://questdb.io/docs/reference/sql/create-table/)
+. This gives you more control over the table schema, allow per-table partitioning, creating indexes, etc.
```
amyshwang, line 219:
```suggestion
In production, it's recommended to use the SQL [CREATE TABLE](https://questdb.io/docs/reference/sql/create-table/) keyword, because it gives you more control over the table schema, allowing per-table partitioning, creating indexes, etc.
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+
+### FAQ
```
amyshwang, line 221:
```suggestion
## FAQ
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+
+**Q**: Does this connector work with Schema Registry?  
+**A**: The Connector does not care about serialization strategy used. It relies on Kafka Connect converters to
```
amyshwang, line 224: I would avoid saying 'does not care about' but not sure if this is correct:

```suggestion
**A**: The Connector works independently of the serialization strategy used. It relies on Kafka Connect converters to
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+deserialize data. Converters can be configured using `key.converter` and `value.converter` options, see the
+configuration section above.
+
+**Q**: I'm getting this
+error: `org.apache.kafka.connect.errors.DataException: JsonConverter with schemas.enable requires "schema" and "payload" fields and may not contain additional fields. If you are trying to deserialize plain JSON data, set schemas.enable=false in your converter configuration.`    
+**A**: This error means that the connector is trying to deserialize data using a converter that expects a schema. The
+connector does not require schemas, so you need to configure the converter to not expect a schema. For example, if you
+are using JSON converter, you need to set `value.converter.schemas.enable=false` or `key.converter.schemas.enable=false`
```
amyshwang, line 232:
```suggestion
connector does not require schemas, so you need to configure the converter to not expect a schema. For example, if you
are using a JSON converter, you need to set `value.converter.schemas.enable=false` or `key.converter.schemas.enable=false`
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+in the connector configuration.
+
+**Q**: Does this connector work with Debezium?  
+**A**: Yes, it's been tested with Debezium as a source and
+a [sample project](https://github.com/questdb/kafka-questdb-connector/tree/main/kafka-questdb-connector-samples/stocks)
+is available. Bear in mind that QuestDB is meant to be used as append-only database hence updates should be translated
```
amyshwang, line 238:
```suggestion
is available. Bear in mind that QuestDB is meant to be used as an append-only database;  hence, updates should be translated
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+as new inserts. The connector supports Debezium's `ExtractNewRecordState` transformation to extract the new state of the
+record. The transform by default drops DELETE events so no need to handle it explicitly.
```
amyshwang, line 240:
```suggestion
record. The transformation by default drops DELETE events, so there is no need to handle them explicitly.
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+
+**Q**: QuestDB is a time-series database, how does it fit into Change Data Capture via Debezium?
```
amyshwang, line 242:
```suggestion
**Q**: QuestDB is a time series database, how does it fit into Change Data Capture via Debezium?  
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+**A**: Yes, it's been tested with Debezium as a source and
+a [sample project](https://github.com/questdb/kafka-questdb-connector/tree/main/kafka-questdb-connector-samples/stocks)
+is available. Bear in mind that QuestDB is meant to be used as append-only database hence updates should be translated
+as new inserts. The connector supports Debezium's `ExtractNewRecordState` transformation to extract the new state of the
+record. The transform by default drops DELETE events so no need to handle it explicitly.
+
+**Q**: QuestDB is a time-series database, how does it fit into Change Data Capture via Debezium?  
+**A**: QuestDB works with Debezium just great! This is the recommended pattern: Transactional applications use a
+relational database to store the current state of the data. QuestDB is used to store the history of changes. Example:
+Imagine you have a Postgres table with the most recent stock prices. Whenever a stock price changes an application
```
amyshwang, line 245:
```suggestion
Imagine you have a PostgreSQL table with the most recent stock prices. Whenever a stock price changes an application
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+updates the Postgres table. Debezium capture each UPDATE/INSERT and pushes it as an event to Kafka. Kafka Connect
```
amyshwang, line 246:
```suggestion
updates the PostgreSQL table. Debezium captures each UPDATE/INSERT and pushes it as an event to Kafka. Kafka Connect
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+QuestDB connector reads the events and inserts them into QuestDB. This way Postgres will have the most recent stock
+prices and QuestDB will have the history of changes. You can use QuestDB to build a dashboard with the most recent stock
```
amyshwang, line 248: Question:
If the most recent prices are in PostgreSQL, then how is it possible to build a price dashboard using QuestDB? 

```suggestion
QuestDB connector reads the events and inserts them into QuestDB. In this way, PostgreSQL will have the most recent stock
prices and QuestDB will have the history of changes. You can use QuestDB to build a dashboard with the most recent stock
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+}
+```
+
+See [ReplaceField documentation](https://docs.confluent.io/platform/current/connect/transforms/replacefield.html#replacefield)
+for more details.
+
+**Q**: I need to run Kafka Connect on Java 8, but the connector says it requires Java 11. What should I do?  
+**A**: The Kafka Connect-specific part of the connectors works with Java 8. The requirement for Java 11 is coming
+from QuestDB client itself. The zip archive contains 2 JARs: `questdb-kafka-connector-VERSION.jar`
+and `questdb-VERSION.jar`. You can use replace the latter with `questdb-VERSION-jdk8.jar` from
```
amyshwang, line 278:
```suggestion
and `questdb-VERSION.jar`. You can replace the latter with `questdb-VERSION-jdk8.jar` from
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+the [Maven central](https://mvnrepository.com/artifact/org.questdb/questdb/6.5.4-jdk8). Bear in mind this setup is not
```
amyshwang, line 279:
```suggestion
the [Maven central](https://mvnrepository.com/artifact/org.questdb/questdb/6.5.4-jdk8). Please note that this setup is not
```

amyshwang (changes requested):
Hi I have gone through `ilp.md`. I only understand the high level concept of Kafka so I worry some suggestions might be wrong - however I try to come up with suggestion so you can just add them to one commit.
One thing I am not sure is the terms. I think for people knowing Kafka well, it does not matter. But for document consistency, I think we should decide on fix ways of describing:
- The connector: I use the Kafka Connection connector or just "the connector" in this page. I like how you introduced the concept at the beginning but then I am dropping the "ILP" piece of info. Is this okay? Maybe we can contextualise the connector with ILP in the intro?
- Kafka: I think it is Apache Kafka? I am not sure but I think we should specify what it is in page and section titles at least.

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+the [reference part.](#reference-documentation)
+
+## Quick Guide
+We start with a practical guide showing how to use the ILP connector to read JSON data from Kafka topic and write them
+as rows into a QuestDB table.
+### Prerequisites
+
+You will need the following:
+
+- Kafka
```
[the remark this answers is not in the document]
jerrinot: Apache Kafka is the full name and trademark. Surely the Apache Foundation (the trademark owner) would prefer to use the full name exclusively. I feel it's too repetitive and the project is colloquially called just "Kafka". I'm good with either option, up to you, dear editor :)

docs/third-party-tools/kafka/ilp.md
jerrinot: I will double-check, but I believe the target table is created even before ILP data are physically committed.

docs/third-party-tools/kafka/ilp.md
jerrinot: you suggestion is correct, thanks!

docs/third-party-tools/kafka/ilp.md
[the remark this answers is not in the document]
amyshwang: Agree we can drop Apache. Reading through this again I wonder if   we should be consistent in using Kafka or Kafka Connect? (Cf L10)

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Publish messages
+
+Messages can be published via the console producer script:
+
+```shell
+bin/kafka-console-producer.sh --topic example-topic --bootstrap-server localhost:9092
+```
+
+A `>` greater-than symbol indicates that a message can be published to the
```
amyshwang, line 99:
```suggestion
A greater-than symbol, `>`, indicates that a message can be published to the
```

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Configuration manual
+This section lists configuration options as well as further information about the Kafka Connect connector.
+### Configuration Optons
+
+The connector supports following configuration options:
+
+| Name                              | Type    | Example                                                     | Default            | Meaning                                                    |
+|-----------------------------------|---------|-------------------------------------------------------------|--------------------|------------------------------------------------------------|
+| topics                            | STRING  | orders                                                      | N/A                | Topics to read from                                        |
```
amyshwang, line 154: I don't think we use capital letters for data type elsewhere. I think `string` would do?

docs/third-party-tools/kafka/ilp.md
amyshwang: TBD this has not been incorporated.

docs/third-party-tools/kafka/ilp.md
```diff
@@ -3,16 +3,16 @@ title: ILP connector
 description: QuestDB ships a Kafka Connect connector for ingesting messages from a Kafka via the ILP protocol.
```
amyshwang, line 3: A Kafka - a Kafka Connect instance?

amyshwang (commented):
Highlighting points to discuss!

docs/third-party-tools/kafka/ilp.md
amyshwang: - "Kafka connector" using Kafka Connect
- Apache to be mentioned in the first instance and then just Kafka
- ILP connector using Kafka Connect --> QuestDB Kafka connector (Kafka Connect connector)

amyshwang:
To summarise our discussion with @jerrinot :

1. Naming ILP connector to QuestDB Kafka connector so we will have:

- QuestDB Kafka connector (using ILP)
- JDBC connector (using JDBC)

They are both base on Kafka Connect. 
2. Mentioning Apache Kafka in the first instance and then it should just be Kafka
3. Kafka Connect is a framework on with both QuestDB Kafka connector and JDBC connector are based

docs/third-party-tools/kafka/ilp.md
amyshwang: TBD as "arrives" is vague - as a user, I don't know if it means the row will be visible to me

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -0,0 +1,320 @@
...
+
+You will need the following:
+
+- Kafka
+- A running QuestDB instance
+
+### Configure Kafka
+
+The Apache Kafka distribution contains the Kafka Connect framework, but to
+complete installing the connector, an archive needs to be downloaded from
```
amyshwang, line 33: I change the wording here because the original gives the impression that Kafka Connect is external to the connector, which I think has changed since our discussion. I don't have a word for "the connector minus Kafka Connect" so this is what I came up with.

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -0,0 +1,320 @@
...
+```shell
+curl -s https://api.github.com/repos/questdb/kafka-questdb-connector/releases/latest | /
+jq -r '.assets[]|select(.content_type == "application/zip")|.browser_download_url'| /
+wget -qi -
+```
+
+:::
+
+A configuration file `/path/to/kafka/config/questdb-connector.properties` must
+be created for Kafka Connect in the standalone mode. The QuestDB host must match
```
amyshwang, line 58: I don't understand what the standalone mode is but I assume users would know! This section (L57-60) is the part that I understand the least, especially "the QuestDB host" and "the destination QuestDB". I think the destination is where the messages will be sent to. I am not sure about the host. I think it is again another concept that users would understand.

amyshwang (commented):
Highlighting some points about my changes and places I am less sure about.

docs/third-party-tools/kafka/questdb-kafka.md
jerrinot: what I wanted to say: the config option "host" must point to the IP address (or host/domain/whatever) where QuestDB is running at. 

does it make sense?

docs/third-party-tools/kafka/questdb-kafka.md
amyshwang: Thanks! Changed parts to components and added this to `dd5a3`

docs/third-party-tools/kafka/questdb-kafka.md
amyshwang: I see! I wonder if this step is before or after a user starts the connector? I think the standalone mode refers to:


> 3. Start the QuestDB Kafka connector:
> 
> ```shell
> bin/connect-standalone.sh config/connect-standalone.properties config/questdb-connector.properties
> ```

So I move this little section on config as step 4. If this is not correct, I'd move it to step 3 and the current step 3 should become step 4.

jerrinot: indeed, first you have to create the `questdb-connector.properties` file and only then you can start the connector.

amyshwang: Understood - reverting this in `0ab998` with bullet points.

docs/third-party-tools/kafka/overview.md
```diff
@@ -0,0 +1,74 @@
...
+
+## QuestDB Kafka connector
+
+QuestDB has developed a
+[QuestDB Kafka connector](/docs/third-party-tools/kafka/questdb-kafka/) for Kafka. The
+connector uses [Kafka Connect](https://kafka.apache.org/documentation/#connect)
+and the QuestDB
+[Influx Line Protocol (ILP)](/docs/develop/insert-data/#influxdb-line-protocol/):
+
+- Kafka Connect: It is part of the Apache Kafka distribution and provides a framework
```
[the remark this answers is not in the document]
amyshwang: It's better indeed - added this to the commit `0ab998`

amyshwang (approved):
Both @jerrinot and I are happy with the content.

docs/third-party-tools/kafka/ilp.md
```diff
@@ -0,0 +1,280 @@
...
+The connector does not do data deserialization on its own. It relies on Kafka Connect converters to deserialize data.
+It's been tested predominantly with JSON, but it should work with any converter, including Avro. Converters can be
+configured using `key.converter` and `value.converter` options, see the table above.
+
+### Designated Timestamps
+
+The connector supports [designated timestamps.](https://questdb.io/docs/concept/designated-timestamp/) If the message
+contains a field with a timestamp, the connector can use it as a timestamp for the row. The field name must be
+configured using `timestamp.field.name` option. The field must either a simple number or a timestamp. When it's a simple
+number, the connector will interpret it as a Unix timestamp in milliseconds.
```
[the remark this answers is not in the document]
jerrinot: this now says:
> When the field is set to an integer the connector will interpret it as a Unix timestamp in milliseconds.

docs/third-party-tools/kafka/ilp.md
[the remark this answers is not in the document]
jerrinot: it should be just Kafka. 
Kafka Connect is part of (Apache) Kafka distribution.

argshook (approved):

PR #1166, opened 2022-11-01
chore(docs) Refining millis and micros unit format

amyshwang:
1. Separating the following:

- `SSS`
- `S`
- `UUU`
- `U`
2. Amending wording for microsecond precision
3. Updating date and timestamp format section
Adding examples in the function sections too.


docs/reference/function/date-time.md
```diff
@@ -949,64 +968,67 @@ Return value type is `int`
...
+| `K`    | Hour in am/pm (0-11)                                                                              | Number             | 0                                     |
+| `h`    | Hour in am/pm (1-12)                                                                              | Number             | 12                                    |
+| `m`    | Minute in hour                                                                                    | Number             | 30                                    |
+| `s`    | Second in minute                                                                                  | Number             | 55                                    |
+| `SSS`  | 3-digit millisecond                                                                               | Number             | 978                                   |
+| `S`    | 3-digit millisecond parsing 1 digit when followed by another `unit`. Otherwise, parsing 3 digits. | Number             | 900                                   |
```
bluestreak01, line 1010: Can we say:

`1- or 3-digit millisecond....` ?

bluestreak01 (changes requested):

docs/reference/function/date-time.md
amyshwang: Updated the text - it is also possible to have 2-digit output so I did not use exactly your wording.

bluestreak01:
Lets not rush things and proof read thoroughly

image

bluestreak01:
```
The date and timestamp format is a combination of units from the table below. These units are case-sensitive
```

I'm not sure why we say "and they used as they are, without any prefix". What prefix is implied?

bluestreak01:
The title should probably be:

```
Date and timestamp format and units
```

We're describing date format, which is a combination of units and text.

amyshwang:
I took the decision to use "letters" in the intro because you cannot really say units are case-sensitive. I will probably rewrite the whole thing...

amyshwang:
I rewrote the intro to:

```
## Date and timestamp formats

Each date or time component is represented by units and arbitrary text. A unit is a combination of letters as defined by the table below. The letters used to form a unit are case-sensitive.
```
I think it is clearer...

bluestreak01 (approved):

PR #1169, opened 2022-11-02
chore(docs) ILP partition default

### LINKED ISSUE #1060: Bug - Misleading partitioning info in ILP

amyshwang:
Closes #1060


docs/concept/partitions.md
```diff
@@ -21,7 +21,7 @@ import Screenshot from "@theme/Screenshot"
 ## Properties
 
 - Available partition intervals are `NONE`, `YEAR`, `MONTH`, `DAY`, and `HOUR`.
-- Default behavior is `PARTITION BY NONE`.
+- Default behavior is `PARTITION BY NONE` using [CREATE TABLE](/docs/reference/sql/create-table) and `PARTITION BY DAY` via [ILP ingestion](/docs/reference/api/ilp/overview/).
```
jerrinot, line 24: when using? the word *when* is missing?

jerrinot (approved):
see the comment, looks good otherwise!

docs/concept/partitions.md
amyshwang:
```suggestion
- Default behavior is `PARTITION BY NONE` when using [CREATE TABLE](/docs/reference/sql/create-table) and `PARTITION BY DAY` via [ILP ingestion](/docs/reference/api/ilp/overview/).
```

jerrinot (approved):

PR #1171, opened 2022-11-03
chore(docs) Improving ILP default timestamp wording

amyshwang:
Adding wording about QuestDB provides microsecond resolution although parsing to nanosecond precision.
Also did some minor editorial changes around the section.


docs/reference/api/ilp/overview.md
```diff
@@ -34,19 +34,19 @@ docs.
...
+| :----------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `table_name` | Name of the table where QuestDB will write data.                                                                                                                                           |
+| `symbolset`  | A set of comma-separated `name=value` pairs that will be parsed as symbol columns.                                                                                                         |
+| `columnset`  | A set of comma-separated `name=value` pairs that will be parsed as non-symbol columns.                                                                                                     |
+| `timestamp`  | UNIX timestamp. The default unit is nanosecond. The value will be truncated to microsecond resolution when parsed by QuestDB. The default can be changed by updating `line.tcp.timestamp`. |
```
jerrinot, line 42: nit:
> The default can be changed by updating [...]
This is ambiguous. The sentence could mean:
1. Default timestamp value
2. Default timestamp units

Perhaps it could be slightly reworded? Something like this:
> UNIX timestamp. The default unit is nanosecond and can be changed by updating `line.tcp.timestamp`. The value will be truncated to microsecond resolution when parsed by QuestDB. 

The passive voice (_[...] can be changed [..]_) is still _meh_, but the ambiguity is gone. 
WDYT?

amyshwang: Good point about the ambiguity and passive voice. I am changing it to:

```
UNIX timestamp. The default unit is nanosecond and is configurable via line.tcp.timestamp. The value will be truncated to microsecond resolution when parsed by QuestDB.
```

jerrinot (approved):

jerrinot (approved):

PR #1172, opened 2022-11-03
chore(docs) Updating `IN` documentation

### LINKED ISSUE #1076: Update - complete documentation for the `IN` operator
**Describe the requested changes:**

The IN operator has some extra features. The documentation suggest we can add an extra modifier using semicolon as the separator to increase/decrease the time range. What it doesn't say is you can actually add extra parameters (not sure what the official name for those is) to repeat the filter in subsequent time periods.

Now with an example (you can run this at https://demo.questdb.io/)

Basic IN works like this and would search anything within that particular day

select pickup_datetime from trips WHERE pickup_datetime IN '2018-06-01T13:00:00'

Now I can extend to search from that timestamp and extend to an extra point in time, let's say 1 minute

select pickup_datetime from trips WHERE pickup_datetime IN '2018-06-01T13:00:00;1m'

So far, so good. All of this was documented. Now imagine I want to search records within the time interval of that minute, but in consecutive days. This is supported with this syntax:

select pickup_datetime from trips WHERE pickup_datetime IN '2018-06-01T13:00:00;1m;24h;10'

Or the equivalent

select pickup_datetime from trips WHERE pickup_datetime IN '2018-06-01T13:00:00;1m;1d;10'

What we are doing here is saying: I want to repeat this filter for every 24h (in the first query) or 1d (in the second query) and do this 10 times. The result of both queries above would be getting all the trips from 13:00 to 13:01 from 2018-06-01 to 2018-06-10.

Let's see what happens If we change the 24h/1d parameter to 12h, like this

select pickup_datetime from trips WHERE pickup_datetime IN '2018-06-01T13:00:00;1m;12h;10'

In this case it would give us results for each interval from 13:00 to 13:01 and also from 01:00 to 01:01, since we are repeating every 12h. Since we are doing this 10 times, we would be getting results for 5 days (10 intervals of 12 hours), finishing on 2018-06-06T01:01:00

Note: This is not supported if the modifier is a negative range

Location:

https://questdb.io/docs/reference/sql/where#time-range

amyshwang:
1. Closes #1076 
2. Move features not yet supported to interval note with a link to the issue for tracking


docs/reference/sql/where.md
```diff
@@ -310,16 +310,25 @@ SELECT * FROM scores WHERE ts IN '2018-05-23T12:15';
...
+![Flow chart showing the syntax of the WHERE clause with a timestamp/modifier comparison](/img/docs/diagrams/whereTimestampIntervalSearch.svg)
 
-`multiplier` is a signed integer.
+- `period` is an unsigned integer.
+  <!--change to signed when this is merged: https://github.com/questdb/questdb/issues/2509-->
 
-- A `positive` value extends the interval.
-- A `negative` value reduces the interval.
+  - A `positive` value extends the selected period.
+  - A `negative` value reduces the selected period.
```
ideoma, line 324: I thought we don't support negative values. Do negative periods work?

docs/reference/sql/where.md
```diff
@@ -346,6 +357,26 @@ Jan 2018) by 3 days.
...
+
+```
+The range is Jan 1 2018 with a one-year interval and the total occurrence is two.
+
+| ts                          | score |
+| --------------------------- | ----- |
+| 2018-01-01T00:00:00.000000Z | 123.4 |
+| ...                         | ...   |
+| 2018-01-01T23:59:59.999999Z | 113.8 |
+| ...                         | ...   |
```
ideoma, line 376: `...` means there is gap. Can be the result have a gap between these timestamps?

docs/reference/sql/where.md
amyshwang: Fixed by `[b2f1bfa](https://github.com/questdb/questdb.io/pull/1172/commits/b2f1bfa93604cdda392d362e99bfda2de72acaa9)`

docs/reference/sql/where.md
amyshwang: From other examples in this section `...` means skipping records. I can change the other examples but I am not sure about the best notation?

ideoma: I beleive there should be no records between `2018-01-01T00:00:00.000000Z` and `2018-01-01T23:59:59.999999Z` and find `...` confusing

amyshwang: I see - sorry I thought you meant all of them. The wrong row has been removed!

ideoma (approved):

argshook (approved):

PR #1177, opened 2022-11-07
chore(docs) `N+` and `U+`

amyshwang:
Capturing
- https://github.com/questdb/questdb/pull/2709
- https://github.com/questdb/questdb/pull/2688


docs/reference/function/date-time.md
```diff
@@ -1015,7 +1015,9 @@ table below. The letters used to form a unit are case-sensitive.
 | `X`    | Time zone                                                                                                      | ISO 8601 time zone | -08; -0800; -08:00                    |
 | `UUU`  | 3-digit microsecond                                                                                            | Number             | 698                                   |
 | `U`    | Microsecond up to 3 digits: `U` parses 1 digit when followed by another `unit`. Otherwise, it parses 3 digits. | Number             | 600                                   |
-| `N`    | Nanosecond                                                                                                     | Number             | 125                                   |
+| `U+`   | 6-digit microsecond                                                                                            | Number             | 600                                   |
+| `N`    | Nanosecond. QuestDB provides microsecond resolution so the parsed nanosecond will be truncated.                | Number             | N/A (truncated)                       |
```
bluestreak01, line 1019: These “units” solve a problem, but when I read description in isolation I cannot relate to anything. This is a specially the case of “+” units that were bourn out of real need

bluestreak01 (changes requested):

docs/reference/function/date-time.md
amyshwang: I am keeping this section slim but linking and adding to the timestamp guide. Will ned to verify the result when 6.5.5 is available either on demo or on Homebrew though.

bluestreak01 (approved):

docs/reference/function/date-time.md
```diff
@@ -1015,7 +1019,9 @@ table below. The letters used to form a unit are case-sensitive.
...
+| `N+`   | 9-digit nanosecond. QuestDB provides microsecond resolution so the parsed nanosecond will be truncated.        | Number             | N/A (truncated)                       |
```
bluestreak01, line 1024: do you think it would make sense to link the section you added with N+ example to this reference manual ?

amyshwang: I did - L989-992. This is a big section so it is a bit far away from this line...

PR #1182, opened 2022-11-10
Add Datron use case

insmac:


src/pages/customers.tsx
```diff
@@ -432,6 +432,36 @@ const Customer = () => {
...
+            <em className={caCss.card__author}>
+              - <strong>Tim Borowski</strong>, DATRON
+            </em>
+            <Button className={caCss.card__cta} to="/case-study/datron/">
+              View full case study
+            </Button>
+          </p>
+        </div>
+      </section>
+
```
amyshwang, line 464: L468, image needs to be on the right-hand side now.

amyshwang (changes requested):
Incorporating changes suggested by @tris0laris

amyshwang (approved):

PR #1180, opened 2022-11-09
feat(docs) INSERT INTO WITH AS 

amyshwang:
- Capturing https://github.com/questdb/questdb/pull/2724
- Removing references to outdated import guide

PR #1187, opened 2022-11-14
feat(docs) existing INSERT INTO WITH AS

amyshwang:
- Capturing https://github.com/questdb/questdb/pull/2724
- Removing references to outdated import guide

This is a temporary update before up coming changes to align to PostgreSQL-compatible behaviour described in https://github.com/questdb/questdb.io/pull/1180


amyshwang:
To work on expected behaviour and its doc.

PR #1192, opened 2022-11-17
chore(docs) python example for PG data insertion

### LINKED ISSUE #999: Update - outdated PG Wire Python example in  develop > insert data

amyshwang:
Closes #999


docs/develop/insert-data.md
```diff
@@ -421,49 +421,44 @@ functionality is on the roadmap for the antecedent
...
-        cursor.close()
-    if connection:
-        connection.close()
-    print('Postgres connection is closed.')
+        print(now)
+        print(date)
+
+        # Insert data into the table.
+
+        for x in range(10):
```
amyshwang, line 445: `        for x in range(10):
            now = dt.now()
            now_timestamp = dt.timestamp(now)
            print(now_timestamp)
            date = dt.today()
            print(date)
            date_timestamp = dt.timestamp(date)

            cur.execute('''
                INSERT INTO test_pg
                    VALUES (%s, %s, %s, %s);
                ''',
                (now_timestamp, date_timestamp, 'python example', x))`

docs/develop/insert-data.md
```diff
@@ -415,55 +415,44 @@ docker run -it --rm --network=host -e PGPASSWORD=quest \
...
-adapter, which does not support prepared statements (bind variables). This
-functionality is on the roadmap for the antecedent
+This example uses the
 [psychopg3](https://github.com/psycopg/psycopg3/projects/1) adapter.
 
 ```python
-import psycopg2 as pg
-import datetime as dt
+import psycopg as pg
+from datetime import datetime as dt
```
amyshwang, line 423: What is the difference between the existing and:
```suggestion
import datetime as dt
```
Sometimes you want to import some of the module?
I have seen:
`from datetime import date`

docs/develop/insert-data.md
```diff
@@ -415,55 +415,44 @@ docker run -it --rm --network=host -e PGPASSWORD=quest \
...
+        # Execute a command: this creates a new table
+        cur.execute('''CREATE TABLE IF NOT EXISTS test_pg (
+        ts TIMESTAMP, name STRING, value INT)
         timestamp(ts);''')
+        print ('Table created.')
+
+        # Insert data into the table.
+
+        for x in range(10):
+            timestamp = int(dt.timestamp(dt.now())*1000000)
```
amyshwang, line 439: This feels very hacky...

docs/develop/insert-data.md
```diff
@@ -415,55 +415,44 @@ docker run -it --rm --network=host -e PGPASSWORD=quest \
...
+            # Converting datetime into millisecond for QuestDB
```
amyshwang, line 440:
```suggestion
            # Convert "datetime" timestamp (double) into millisecond for QuestDB
```

docs/develop/insert-data.md
amyshwang: p.s. `import datetime as dt` causes this error:
```
➜  python_amy python3 pg.py
Table created.
Traceback (most recent call last):
  File "/Users/work/Repos/python_amy/pg.py", line 18, in <module>
    timestamp = int(dt.timestamp(dt.now())*1000000)
AttributeError: module 'datetime' has no attribute 'timestamp'
````

File: docs/develop/insert-data.md [hunk 5]
```diff
@@ -415,55 +415,51 @@ docker run -it --rm --network=host -e PGPASSWORD=quest \
 <TabItem value="python">
 
 
-This example uses the [psycopg2](https://github.com/psycopg/psycopg2) database
-adapter, which does not support prepared statements (bind variables). This
-functionality is on the roadmap for the antecedent
+This example uses the
 [psychopg3](https://github.com/psycopg/psycopg3/projects/1) adapter.
```
amunra, line 419: Swap out link for: https://www.psycopg.org/psycopg3/docs/

amunra: Add:
-------


To [install](https://www.psycopg.org/psycopg3/docs/basic/install.html) the client library, use `pip`:

```bash
python3 -m pip install "psycopg[binary]"
```

amunra (changes requested):
Looks good.
A few touch-ups required.

docs/develop/insert-data.md
amunra: > p.s. `import datetime as dt` causes this error:
> 
> ```
> ➜  python_amy python3 pg.py
> Table created.
> Traceback (most recent call last):
>   File "/Users/work/Repos/python_amy/pg.py", line 18, in <module>
>     timestamp = int(dt.timestamp(dt.now())*1000000)
> AttributeError: module 'datetime' has no attribute 'timestamp'
> ```

**TL;DR**: All code that previously read `dt.SOMETHING` should now read `dt.datetime.SOMETHING`.

Here's the why:

```python
>>> # this imports the _module_. It contains classes and functions to work with dates, times, intervals etc.
>>> import datetime
>>> datetime
<module 'datetime' from '/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/datetime.py'>

>>> # the suggested change also imports the module, but aliased.
>>> import datetime as dt
>>> dt
<module 'datetime' from '/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/datetime.py'>

>>> # these really are the same thing and the alias just reduces keystrokes.
>>> dt is datetime
True

>>> # there's also a `datetime.datetime` class which represents a point in time.
>>> datetime.datetime
<class 'datetime.datetime'>
>>> # or, using our alias (which I think is actually clearer in this specific case).
>>> dt.datetime
<class 'datetime.datetime'>

>>> # if we create an instance of it with the current time (UTC timezone)
>>> utc_now = dt.datetime.utcnow()
>>> utc_now
datetime.datetime(2022, 12, 1, 10, 5, 22, 825853)

>>> # We can now convert it to a timestamp
>>> utc_now.timestamp()
1669889122.825853
```

docs/develop/insert-data.md
```diff
@@ -415,55 +415,66 @@ docker run -it --rm --network=host -e PGPASSWORD=quest \
...
+        
+        print('Table created.')
+
+        # Insert data into the table.
+
+        for x in range(10):
+
+            # Converting datetime into millisecond for QuestDB
+
+            timestamp = int(dt.timestamp(dt.now())*1000000)
```
amunra, line 459: this should probably obtain the current `utc` time and not the current local time:

```
timestamp = int(dt.datetime.utcnow().timestamp() * SEC_TO_MICROS)
```

and there should be a global set (right after the import statements) with:

```
SEC_TO_MICROS = 1000000
```

It's worth doing a quick check to be sure the date that appears in the DB is the same as the one you expect and isn't skewed by a few hours because of timezone differences.

amunra: Scratch that.. there's actually a more accurate way to this altogether:

```python
import time

...

timestamp = time.time_ns() // 1000
```

This approach avoids getting confused with timezones, intermediate objects and float to integer conversions.

^^^^^ Do this ^^^^^

It probably means you can drop the `import datetime` part as well.

docs/develop/insert-data.md
amunra: Scratch this... there's a better way to do this with `time.time_ns()`. See comment later on. No need to `import datetime`.

Issue #1188, opened 2022-11-15
Change in commit lag


amyshwang:
https://github.com/questdb/questdb/pull/2759

The current `commitLag` becomes `o3MaxLag` - this plays a role of resource usage limit and one of the controls for automated lag adjuster. It is set to 10 min and users do not need to change it or even know this setting exists.

amyshwang:
1. Terms
- Change commit lag to O3 lag, which is more suitable.
- "automated O3 lag adjuster"
2. Config setting
 - commitLag becomes o3MaxLag
 - config property cairo.commit.lag is left intact for backward compatibility, although we prefer cairo.o3.max.lag if it exists
 - TBD query-wide commit lat setting 

## Relevant pages

PR #1197, opened 2022-11-18
feat(docs) Aligning with commit lag change for 6.6

amyshwang:
Closes [#1188](https://github.com/questdb/questdb.io/issues/1188)

1. Adding notes for sections and pages dedicated to commit lag, moving them to the end of the sidebar section or the end of a page
2. Adding diagrams without the commit lag parameters
3. Remove `commitLag` from examples where it is not relevant
4. Adding the new config to the configuration table


docs/reference/sql/alter-table-set-param.md
```diff
@@ -25,7 +30,8 @@ functions which are described in the
 
 The following two sections describe table parameters relating to out-of-order
 ingestion. For context on commit lag and max uncommitted rows, see the guide for
-[configuring commit lag of out-of-order data](/docs/guides/out-of-order-commit-lag) and [ILP commit strategy](/docs/reference/api/ilp/tcp-receiver#commit-strategy).
+[configuring commit lag of out-of-order data](/docs/guides/out-of-order-commit-lag)
```
bluestreak01, line 33: we need to describe the meaning of o3MaxLag here:

`o3MaxLag` defines the max size of in-memory buffer in millis. Buffer is allocated dynamically thru analysing shape of the incoming data with `o3MaxLag` being the upper limit.

docs/reference/sql/insert.md
```diff
@@ -113,19 +94,46 @@ INSERT INTO confirmed_trades
 SELECT * FROM confirmed_id;
 ```
 
-Inserting out-of-order data into an ordered dataset may be optimized using
-`batch` and `commitLag` parameters:
+## Parameters for QuestDB 6.5.5 and earlier versions
```
bluestreak01, line 97: `batch` parameter is still relevant and should be documented

docs/reference/sql/select.md
```diff
@@ -13,14 +13,15 @@ and evaluated from a table.
 
 :::tip
 
-- While ingesting data using ILP, data may not be immediately available to a
-  `SELECT` query, due to the [commit lag](/docs/guides/out-of-order-commit-lag)
-  setting. Please see the
+- For QuestDB 6.5.5 and earlier versions, data ingested via ILP may not be
```
bluestreak01, line 16: lets remove any mention of lag from `SELECT` doc, it is highly confusing

docs/troubleshooting/faq.md
```diff
@@ -10,13 +10,15 @@ issues met when running QuestDB, as well as solutions to them.
 
 InfluxDB line protocol (ILP) does not commit data on single lines or when the
 sender disconnects, but instead uses a number of rules to break incoming data
-into commit batches. This results in data not being visible in `SELECT` queries
-immediately after being received. Refer to our
-[guide to commit lag](/docs/guides/out-of-order-commit-lag) to understand the
-concept and
+into commit batches. Fos QuestDB 6.5.5 and earlier versions, this results in
```
bluestreak01, line 13: this is a confusing statement. I would remove the entire paragraph. What is `false` here is that server does actually commit on disconnect.

docs/troubleshooting/faq.md
```diff
@@ -10,13 +10,15 @@ issues met when running QuestDB, as well as solutions to them.
...
+data not being visible in `SELECT` queries immediately after being received.
```
bluestreak01, line 14: lets not mess with explaining immediacy, data is immediate - should be non-issue for the user. If data is not immediate - upgrade to 6.6.

sidebars.js
```diff
@@ -56,11 +56,11 @@ module.exports = {
       label: "Guides",
       type: "category",
       items: [
-        "guides/out-of-order-commit-lag",
         "guides/importing-data",
         "guides/importing-data-rest",
         "guides/modifying-data",
         "guides/working-with-timestamps-timezones",
+        "guides/out-of-order-commit-lag",
```
bluestreak01, line 63: remove

bluestreak01 (changes requested):

docs/reference/sql/alter-table-set-param.md
amyshwang: Wording added to the configuration section.
If we add the explanation of o3MaxLag here, we might also want to use it in the diagram and the rest of the page?

amyshwang: Marking "Deprecated content" at the top of the page.

docs/concept/designated-timestamp.md
```diff
@@ -31,9 +35,9 @@ Checking if tables contain a designated timestamp column can be done via the
 
 ## Out-of-order policy
```
jerrinot, line 36: maybe we can remove this box? 
image

nobody should be using version before 6.0.

sidebars.js
```diff
@@ -56,7 +56,6 @@ module.exports = {
       label: "Guides",
       type: "category",
       items: [
-        "guides/out-of-order-commit-lag",
```
jerrinot, line 59: I believe we removed it from the menu, because it's no longer relevant and we do not want to bother users with details required for old versions only. and at the same time the page is not entirely removed, because it's indexed by Google? Is that right? If I go to https://deploy-preview-1197--unruffled-blackwell-31bfb2.netlify.app/docs/guides/out-of-order-commit-lag/ then I see the page, but it looks a bit lost. Like it's not part of the docs. Is that intentional?

image

docs/reference/configuration.md
```diff
@@ -471,7 +472,7 @@ line protocol.
...
 | line.tcp.commit.interval.fraction          | 0.5          | Commit lag fraction. Used to calculate commit interval for the table according to the following formula: `commit_interval = commit_lag ∗ fraction`. The calculated commit interval defines how long uncommitted data will need to remain uncommitted. |
-| line.tcp.commit.interval.default           | 2000         | Default commit interval in milliseconds. Used when no commit lag set for a table or the fraction is set to 0.                                                                                                                                         |
+| line.tcp.commit.interval.default           | 1000         | Default commit interval in milliseconds. Used when no commit lag set for a table or the fraction is set to 0.                                                                                                                                         |
```
jerrinot, line 475: > Used when no commit lag set [...]
What does it mean? There is no commit lag configuration anymore. Is it when max.lag is set to 0?

docs/reference/function/meta.md
```diff
@@ -91,24 +84,24 @@ Returns a `table`.
 tables();
 ```
 
-| id  | name        | designatedTimestamp | partitionBy | maxUncommittedRows | commitLag |
-| --- | ----------- | ------------------- | ----------- | ------------------ | --------- |
-| 1   | my_table    | ts                  | DAY         | 500000             | 300000000 |
-| 2   | device_data | null                | NONE        | 10000              | 30000000  |
+| id  | name        | designatedTimestamp | partitionBy | maxUncommittedRows |
```
jerrinot, line 87: it's missing the `o3MaxLag` column. (also the walEnabled, but I am not sure if we want to talk about WAL yet)

jerrinot: image

docs/reference/sql/create-table.md
```diff
@@ -218,6 +218,15 @@ created.
 
 ## WITH table parameters
 
+:::note
```
jerrinot, line 221: I'm a bit confused here. specifically the `maxUncommittedRows` parameter. I don't know how the new commit works, but `tables()` in 6.6 still shows the `maxUncommittedRows` hence I presume it's configurable per-table? While this deprecation notice makes the impression it's gone for good?

docs/troubleshooting/faq.md
```diff
@@ -8,14 +8,8 @@ issues met when running QuestDB, as well as solutions to them.
...
 
-InfluxDB line protocol (ILP) does not commit data on single lines or when the
-sender disconnects, but instead uses a number of rules to break incoming data
-into commit batches. This results in data not being visible in `SELECT` queries
-immediately after being received. Refer to our
-[guide to commit lag](/docs/guides/out-of-order-commit-lag) to understand the
-concept and
-[InfluxDB line protocol reference](/docs/reference/api/ilp/tcp-receiver#commit-strategy)
-to understand these rules.
+Please upgrade to QuestDB 6.6 or later versions for dynamic data ingestion
```
jerrinot, line 11: Even in 6.6 it's not available immediately. It's just the configuration is automatic. but I assume for heavily O3 data the dynamic lag will grow -> data will not be visible -> users will still be confused?

jerrinot (changes requested):
I don't really understand the new impl, but here are the 2 most important notes:
1. there is a confusion about the `maxUncommitedRows` configuration. Is it really gone for good? Why is `tables()` showing it?
2. I think we should not pretend the O3 lag is gone and the problem is solved. There will still be users who won't be confused why there don't see data their inserted. I assume the new impl makes life much easier for users without O3 inserts. But users will O3 will still experience the delay and we should describe it.

sidebars.js
amyshwang: Vlad suggested removing it from the sidebar - as you said, it is irrelevant and we don't want people to land on it unless they have to.
However, I did not notice the behaviour so I will add it back.

docs/reference/configuration.md
amyshwang: I think the system uses it when commit lag (which is dynamic between max.lag and min.lag) value is 0. I will remove this...

docs/reference/function/meta.md
amyshwang: Reverted the deletion and added it back as o3MaxLag

docs/reference/sql/create-table.md
amyshwang: My misunderstanding - only commitLag is gone.

docs/troubleshooting/faq.md
amyshwang: Yes but in this case it is the nature of the data. I feel the answer is not really adding much. I suggest deleting this and collect new FAQ later if required.

jerrinot (approved):

insmac (approved):

docs/guides/out-of-order-commit-lag.md
```diff
@@ -39,54 +52,70 @@ If any new timestamp value has a high probability to arrive within 10 seconds of
 the previously received value, the boundary for this data is `10 seconds` and we
 name this **commit lag** or just **lag**.
 
-Data received within the commit lag value is kept in memory only, and invisible from queries. Our out-of-order algorithm detects and prioritizes the data using an optimized processing path to commit later. Once committed, the data is visible for any relevant queries. 
+Data received within the commit lag value is kept in memory only, and invisible
+from queries. Our out-of-order algorithm detects and prioritizes the data using
+an optimized processing path to commit later. Once committed, the data is
+visible for any relevant queries.
```
tris0laris, line 58: "any relevant" : i would remove

docs/introduction.md
```diff
@@ -21,8 +21,10 @@ methods:
 - [Binaries](/docs/get-started/binaries/) for direct downloads to run on Linux,
   macOS or Windows
 - [Homebrew](/docs/get-started/homebrew/) for running QuestDB on macOS
-- [Create your first database](/docs/get-started/first-database/) provides a guide for once QuestDB is running
-- [Learn more](/docs/get-started/learn-more/) is a selection of useful concepts and guides for new QuestDB users
+- [Create your first database](/docs/get-started/first-database/) provides a
+  guide for once QuestDB is running
```
tris0laris, line 25: I think we need to delete "for" ?

docs/introduction.md
```diff
@@ -76,10 +78,10 @@ This section describes how to integrate QuestDB with third-party tools and
...
 
 - [Prometheus](/docs/third-party-tools/prometheus) monitoring and alerting.
-- [Grafana](/docs/third-party-tools/grafana) instructions for connecting
-  QuestDB as a datasource for building visualizations and dashboards
-- [Kafka](/docs/third-party-tools/kafka/overview) guide for ingesting data from topics
-  into QuestDB
+- [Grafana](/docs/third-party-tools/grafana) instructions for connecting QuestDB
+  as a datasource for building visualizations and dashboards
+- [Kafka](/docs/third-party-tools/kafka/overview) guide for ingesting data from
+  topics into QuestDB
```
tris0laris, line 84: from topics or from Kafka topics?

docs/introduction.md
```diff
@@ -95,8 +97,8 @@ data, and introduces features and capabilities unique to the system.
 - [SQL extensions](/docs/concept/sql-extensions) allow performant time series
   analysis with a concise syntax
 - [JIT compiler](/docs/concept/jit-compiler) to speed up queries
```
tris0laris, line 99: should it not be "WHERE" queries specifically?

docs/introduction.md
```diff
@@ -95,8 +97,8 @@ data, and introduces features and capabilities unique to the system.
...
-- [Partitions](/docs/concept/partitions) can be used for significant
-  performance benefits on calculations and queries
+- [Partitions](/docs/concept/partitions) can be used for significant performance
+  benefits on calculations and queries
```
tris0laris, line 101: this is a bit convoluted. what about: 
should be used to lift less data for time-based queries

docs/guides/out-of-order-commit-lag.md
amyshwang:
```suggestion
visible for queries.
```

docs/introduction.md
amyshwang:
```suggestion
  guide once QuestDB is running
```

docs/introduction.md
amyshwang, line 101:
```suggestion
- [Partitions](/docs/concept/partitions) can be used to perform time-based queries more efficiently
```

amyshwang: I find "list less data" very vague but agree it could be simplified -edited your suggestion. I will keep "can" too because it is a suggestion.

docs/introduction.md
amyshwang: Sorry I don't understand what you meant, this is reflecting the sidebar so the items are fixed?

tris0laris: - [JIT compiler](/docs/concept/jit-compiler) to speed up queries

my question is: should we be more specific, saying that we speed "WHERE" queries ? its a question btw, not asking you to do it

amyshwang:
```suggestion
- [JIT compiler](/docs/concept/jit-compiler) to speed up `WHERE` clauses
```

tris0laris (approved):

PR #1208, opened 2022-11-24
Add missing FreeBSD tile to Download QuestDB page

nrichers:
While testing the docs for https://github.com/questdb/questdb.io/pull/1193, I noticed that your [Download QuestDB](https://questdb.io/get-questdb/) page is missing a tile for FreeBSD. This looks like an omission as all the necessary assets are readily available, so here's a quick fix. 

Output after the fix:

image

Added bonus for those of us who might be slightly neurodivergent: you now have a nice 3x3 grid for your downloads page! 😄 

(@amyshwang here's the fix for you.)


amyshwang (approved):
Thank you!

PR #1214, opened 2022-11-28
chore(docs): include a timestamp configuration option for the Kafka C…

jerrinot:
…onnector


docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -229,8 +230,10 @@ The connector supports
 If the message contains a timestamp field, the connector can use it as a
 timestamp for the row. The field name must be configured using the
 `timestamp.field.name` option. The field must either be an integer or a
-timestamp. When the field is set to an integer the connector will interpret it
-as a Unix timestamp in milliseconds.
+timestamp. When the field is set to an integer the connector will autodetect its units. 
+This works for timestamps after 4/26/1970, 5:46:40 PM. The units can be also configured
+explicitly using the `timestamp.units` configuration option. Supported values are `nanos`,
+`micros`, `millis` and `auto`.
```
amyshwang, line 236:
```suggestion
timestamp. When the field is set to an integer, the connector will autodetect its units. 
This works for timestamps after 04/26/1970, 5:46:40 PM. The units can also be configured
explicitly using the `timestamp.units` configuration, which supports the following values:
- `nanos`
- `micros`
- `millis`
-  `auto` (default)
```

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -182,6 +182,7 @@ The connector supports the following configuration options:
 | value.prefix                      | `string`  | from_value                                                  | N/A                | Prefix for value fields                                    |
 | <sub>skip.unsupported.types</sub> | `boolean` | false                                                       | false              | Skip unsupported types                                     |
 | <sub>timestamp.field.name</sub>   | `string`  | pickup_time                                                 | N/A                | Designated timestamp field name                            |
+| timestamp.units                   | `string`  | micros                                                      | auto               | Designated timestamp field units                           |
```
amyshwang, line 185: Should it be `auto` or auto? Cf L236.

amyshwang (changes requested):
Minor formatting comments!

docs/third-party-tools/kafka/questdb-kafka.md
jerrinot: I gues auto? as it is. the other values in the same column are not back-ticked.

amyshwang (approved):

PR #1216, opened 2022-11-30
blog: the power of sequental scanning

jerrinot:


jerrinot:
@amyshwang I decided to keep it simple. the post is not that good to justify spending too much energy/time on it. it was meant to be a quickie. the only part I am not sure about: the image copied from the original article - we give credits. but still I am not sure if that's ok. without knowing what is the license, etc.

amyshwang (approved):
Thank you!

PR #1219, opened 2022-12-02
chore(docs) update `SET PARA` to move commitLag to the end of the page

amyshwang:
- Max uncommitted rows is still a valid setting
- Commit lag is deprecated after 6.6


docs/reference/sql/alter-table-set-param.md
```diff
@@ -8,6 +8,49 @@ description: SET PARAM SQL keyword reference documentation.
...
+  documentation page.
+
+:::
+
+## Syntax
+
+![Flow chart showing the syntax of the ALTER TABLE keyword](/img/docs/diagrams/alterTable.svg)
+![Flow chart showing the syntax of the ALTER TABLE SET PARA keywords](/img/docs/diagrams/alterTableSetParam.svg)
+
+`maxUncommittedRows` - equivalent to `cairo.max.uncommitted.rows`. The parameter allows for specifying the maximum number of uncommitted
```
jerrinot, line 22: > maxUncommittedRows` - equivalent to `cairo.max.uncommitted.rows`.

this assumes a reader knows what `cairo.max.uncommitted.rows` is. That's not good, especially given it's a first sentence (apart from the NOTE box). Moreover, it's not an equivalent. `cairo.max.uncommitted.rows` is global while this param is per-table. I think this paragraph should be restructured a bit. A suggestion: Start with what is it good for and only then point to the `cairo.max.uncommitted.rows` property.

amyshwang: Makes sense - I updated it but since the original text was copied from CREATE TABLE, I updated CREATE TABLE too!

jerrinot (approved):

PR #1221, opened 2022-12-02
feat(docs) `WAL` documentation

### LINKED ISSUE #1146: Update - Docs update for `WAL`
## What is `WAL`

- WAL (write ahead log) is an alternative write approach to the existing one. 
- WAL is a necessary first step for basic replication. 
- WAL is optional
- WAL achieves:
   - Decoupling data ingestion and O3 writes
   - Isolating data ingestion from other operations
   - Permitting different data ingestion interfaces

This feature implements a dedicated writer for each data ingestion channel (REST API, ILP, PGWire) (`wal writer`) writing to dedicated log (WAL). 

#### WAL

Dedicated API for each ingestion interface. A WAL owns a directory on the disk where column and metadata are saved.

_Properties_

- WAL organises data into segments.
- Segments are numbered, the WAL assigns ids starting with 0 and increments it by 1 for each new segment after that.
- A new segment is started when:

- the storage size of a segment reaches a limit set in configuration
- the number of rows in a segment reaches a limit set in configuration
- after some time set in configuration elapsed
- the schema of the table changes (add/remove columns)
- a row is cancelled

#### Sequencer
- A `sequencer`: centrally manage transactions (single source of truth). The Sequencer also keeps a log where it records which WAL was given which txn numbers and which segment's event file has the details of the transaction.
This log is called `txn catalog` and stored in a meta file called `_catalog`.

    - to provide unique transaction ids
    - to own and manage the table's schema
    - to give out write ahead log (WAL) instances with the unique id
    - maintain the transaction catalog, a log which records which WAL owns which transactions

#### `WalTableWriter`

A 'WAL application job` picks up the requested commits and applies them to the table, following the sequence of unique IDs issued by the `sequencer`.
The 'WAL application job` then sends commits to the `WalTableWriter`, which then processes any O3 writes.

## Changes to the existing mechanism

Currently the TableWriter is owned (locked) by a single thread writing to the files therefore write access is limited. As a table is receiving new data, the following are not permitted:
- Ingestion via another interface
- Table modification
- Data modification

## Configurations

- Config about WAL segments
- Table-wide config **(set when creating tables?)**

**Docs location:**
1. Concept/guide about WAL
A comparison of WAL and non-WAL tables 

Document WAL table quirks 

- Nature of async operations application, e.g. that SELECTs may not see changes INSERTs, UPDATEs or ALTERs executed just before them
- Limitations that UPDATE on WAL table do not support JOINs
- UPDATE does not return back the count of updated rows
- Various data related ALTER table like drop / attach partition may fail silently **when target partitions don't exist** [Data dependent failures are not guaranteed]
- ALTER TABLE .. ADD COLUMN can only add 1 column per statement
- No Table Busy writing from multiple connections supported across all interfaces
- Immediate DROP and RENAME of the table without system-wide lock.
- Decoupling folder names and the table names (system table name) - To implement for detach and attach (TBD)

3. Configurations
default - CAIRO_WAL_ENABLED_DEFAULT, false
5. -Commit lag relevant concepts may need a note pointing to how WAL changes the operation- Done by 6.6
6. Where `TableWriter` is mentioned: to check whether updates are required:- https://questdb.io/docs/concept/geohashes/#java-embedded-usage
- https://questdb.io/docs/reference/api/java-embedded/
- https://questdb.io/docs/concept/root-directory-structure/
6. Individual SQL keywords:
- CREATE TABLE

```
CREATE TABLE Abc (i INT, ts timestamp) 
timestamp(ts)
partition by DAY
WAL

CREATE TABLE Abc (i INT, ts timestamp) 
timestamp(ts)
partition by DAY
BYPASS WAL
```
- ALTER TABLE
- UPDATE
- RENAME
- DROP
- VACUUM
- TRUNCATE (TBD if implemented)

## Relevant PRs
- [First PR with overall description](https://github.com/questdb/questdb/pull/2144)

- https://github.com/questdb/questdb/pull/2335
Keyword writeMode=WAL/DIRECT added (Cf CREATE TABLE WAL BYPASS)?
- https://github.com/questdb/questdb/pull/2457
- https://github.com/questdb/questdb/pull/2665
- https://github.com/questdb/questdb/pull/2752

PR enabling monitoring of WAL table - 
https://github.com/questdb/questdb/pull/2841/files

amyshwang:
- A new page (concept) on WAL v.s. non-WAL
- Individual SQL pages to be updated

closes #1146


docs/concept/write-ahead-log.md
```diff
@@ -0,0 +1,63 @@
...
+A WAL table provides the following traits:
+
+- WAL decouples data ingestion from other operations.
+- The use of WAL permits concurrent data ingestion using different interfaces.
+
+The following are WAL-related settings:
+
+- It is optional to enable WAL to a table
+- WAL is set when a table is created. See [`CREATE TABLE`](docs/reference/sql/create-table/) for more information.
+- The server-wide setting to enable WAL is `CAIRO_WAL_ENABLED_DEFAULT`.
```
amyshwang, line 23: This is probably not something users need to know...

docs/concept/geohashes.md
```diff
@@ -296,14 +296,30 @@ For more information on the use of this operator, see the
...
+Geohashes are inserted into tables via Java (embedded) QuestDB instance through
+the selected `Writer`'s `putGeoHash` method. The `putGeoHash` method accepts
+`LONG` values natively with the destination precision. Additionally,
+`GeoHashes.fromString` may be used for string conversion, but comes with some
+performance overhead as opposed to `long` values directly.
+
+The appropriate `Writer` to use depends on whether the table is a
+[WAL](docs/concept/write-ahead-log/) table or not:
+
+- `TableWriter` is used for a non-WAL table.
```
glasstiger, line 308: i would say that:
- `TableWriter` is used to write data directly into a table
- `WalWriter` is used to write data into a WAL-enabled table via WAL

That is because via embedded API you could just get a TableWriter for a WAL-enabled table and start adding rows.
When using embedded API you have much more freedom than if you were coming through ILP.

About the 3rd option, I would just mention that the different Writers implement a common interface called `TableWriterAPI` and the embedded API provides an option to request a `Writer` with the interface type. This API provides the suitable `Writer` instance by checking if the table is WAL-enabled or not.

docs/concept/write-ahead-log.md
```diff
@@ -0,0 +1,148 @@
...
+  numbers and send them them to the `WalWriter` to be committed.
+
+- `WalWriter`: updates the database and resolves any out-of-order data
+  writes.
+
+<Screenshot
+  alt="Diagram showing the sequencer allocating txn numbers to events cronologically"
+  title="The sequencer allocates unique txn numbers to events from different WALs chronologically and serves as the single source of truth."
+  height={435}
+  src="/img/docs/concepts/wal_sequencer.png"
```
glasstiger, line 135: on this pic "an event making commit requests" could be just one of these:
- an event
- a commit request
- a transaction

i would probably use the last one, all these events are transactions which will need a commit.
in fact the commit already happened and the transaction was created when the data was accepted by the WAL and the Sequencer provided a txn number for it.
we are just passing these transactions to the WAL application process as events in a queue (txn 1, txn 2, txn 3 and so on...).

docs/concept/write-ahead-log.md
```diff
@@ -0,0 +1,148 @@
...
+  │   │   ├── _meta
+  │   │   ├── _txn
+  │   │   └── _cv
+  │   |
+  ```
+
+- `WAL application job`: collects the commit requests based on the unique `txn`
+  numbers and send them them to the `WalWriter` to be committed.
+
+- `WalWriter`: updates the database and resolves any out-of-order data
```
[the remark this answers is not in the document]
amyshwang: Is this the same `TableWriter` used for non-WAL table? I thought this is the main new component. Same question for the diagram where I use `WalWriter`.

docs/concept/geohashes.md
amyshwang: I am drafting something like this:

> The following configurations enable WAL tables:

> - WAL is an optional feature that is disabled by default, as set by `cairo.wal.supported`.
> - When `cairo.wal.supported` is swtiched on , WAL tables can be created using one of the following settings:
> 
>   - Table-wide configuration [`CREATE TABLE...WAL`](docs/reference/sql/create-table/) and [`CREATE TABLE...BY PASS`]: `WAL` generates a WAL table and `BY PASS` generate a non-WAL table.
>   - Server-wide configuration `cairo.wal.enabled.default`: When set to `true`, [`CREATE TABLE`](docs/reference/sql/create-table/) SQL keyword generate WAL-tables when `WAL` and `BY PASS` are not used.

1. But I wonder if we really want to encourage users to change the server-side default? 
2. I can only find `CAIRO_WAL_ENABLED_DEFAULT("cairo.wal.enabled.default")` in the repo

docs/concept/write-ahead-log.md
amyshwang: I am drafting something like this:

> The following configurations enable WAL tables:

> - WAL is an optional feature that is disabled by default, as set by `cairo.wal.supported`.
> - When `cairo.wal.supported` is swtiched on , WAL tables can be created using one of the following settings:
> 
>   - Table-wide configuration [`CREATE TABLE...WAL`](docs/reference/sql/create-table/) and [`CREATE TABLE...BY PASS`]: `WAL` generates a WAL table and `BY PASS` generate a non-WAL table.
>   - Server-wide configuration `cairo.wal.enabled.default`: When set to `true`, [`CREATE TABLE`](docs/reference/sql/create-table/) SQL keyword generate WAL-tables when `WAL` and `BY PASS` are not used.

1. But I wonder if we really want to encourage users to change the server-side default? 
2. I can only find `CAIRO_WAL_ENABLED_DEFAULT("cairo.wal.enabled.default")` in the repo

docs/concept/write-ahead-log.md
```diff
@@ -0,0 +1,148 @@
...
+From QuestDB TBD, we add a new approach to ingest data using a write-ahead log
+(WAL). This page introduces the properties of a WAL-enabled table (WAL table)
+and compares it with a non-WAL table.
+
+## Properties
+
+A WAL table provides the following traits:
+
+- WAL decouples data ingestion from other operations.
+- The use of WAL permits concurrent data ingestion using different interfaces.
```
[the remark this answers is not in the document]
amyshwang: and schema changes

docs/reference/api/java-embedded.md
```diff
@@ -89,7 +89,63 @@ import TabItem from "@theme/TabItem"
...
+    try (SqlCompiler compiler = new SqlCompiler(engine)) {
+        compiler.compile("create table testTable (" +
+                "a int, b byte, c short, d long, e float, g double, h date, " +
+                "i symbol, j string, k boolean, l geohash(8c), ts timestamp" +
+                ") timestamp(ts) partition by day WAL", ctx);
+
+        // write data into WAL
+        try (WalWriter writer = engine.getWalWriter(ctx.getCairoSecurityContext(), "testTable")) {
+            for (int i = 0; i < 3; i++) {
+                TableWriter.Row row = writer.newRow(Os.currentTimeMicros());
```
amyshwang, line 121: I wonder if this should be:

```suggestion
                WalWriter.Row row = writer.newRow(Os.currentTimeMicros());
```

amyshwang: Actually `TableWriter` is used in "New instance", "Create a new row" I wrote code examples for `WalWriter` and `TableWriterAPI` but I am not sure if this is correct.

docs/reference/api/java-embedded.md
```diff
@@ -219,43 +274,64 @@ always be 1.
 
 #### New SqlCompiler instance and blank table
 
-Before we start writing data using `TableWriter`, the target table has to exist.
-There are several ways to create new table ; using `SqlCompiler` is the easiest.
+Before we start writing data using a writer, the target table has to exist. There are several
+ways to create a new table and we recommend using `SqlCompiler`:
 
 ```java title="Creating new table"
+
```
amyshwang, line 281: Here, I don't think `TableWriterAPI` would work because one should know what is to be created.

docs/reference/api/java-embedded.md
```diff
@@ -219,43 +274,64 @@ always be 1.
...
+try (WalWriter writer = engine.getWriter(ctx.getCairoSecurityContext(), "abc", "testing")) {
+```
+
+```java title="New table writer instance for either a WAL or non-WAL table"
+
+try (TableWriterAPI writer = engine.getWriter(ctx.getCairoSecurityContext(), "abc", "testing")) {
+
+```
+
+A non-WAL table uses `TableWriter`, which will hold exclusive lock on table `abc` until it is closed and
```
amyshwang, line 318: TBD what to say about `WalWriter`. It does not hold an exclusive lock?

docs/reference/api/java-embedded.md
```diff
@@ -219,43 +274,64 @@ always be 1.
...
 inter-process. If you have two Java applications accessing the same table only
 one will succeed at one time.
 
+
 #### Create a new row
 
+
 ```java title="Creating new table row with timestamp"
 TableWriter.Row row = writer.newRow(Os.currentTimeMicros());
 ```
```
amyshwang, line 329: I think we should add examples for the two other writers as well.

docs/concept/write-ahead-log.md
glasstiger: "a transaction making commit requests" should be just "transaction".
Transactions are the result of commits (or you could say commit requests).

I think the confusion comes from the timing of these events.
First a `commit` happens which results in a `transaction` and an `event` is passed to the WAL apply job so it knows that there is a new transaction in the WAL which needs to be applied to the table. These happen at the same time basically.

docs/reference/api/java-embedded.md
glasstiger: this is fine, there is no WalWriter.Row.
we actually talked about it some time ago that we should move Row out of TableWriter into a separate class, maybe called TableRow or just Row.
we can still do it but for now it is TableWriter.Row

docs/reference/api/java-embedded.md
```diff
@@ -89,14 +89,76 @@ import TabItem from "@theme/TabItem"
...
+
+Non-WAL tables do not allow concurrent writes via multiple interfaces. To
+successfully create an instance, the table must:
 
 - Already exist
 - Have no other open writers against it as the `TableWriter` constructor will
   attempt to obtain an exclusive cross-process lock on the table.
 
-```java title="Example table writer"
+```java title="Example TableWriter"
```
glasstiger, line 161: i would update this example to the one provided yesterday just to make it use the same table name and formatting as the other code examples

docs/reference/api/java-embedded.md
```diff
@@ -126,31 +188,79 @@ try (CairoEngine engine = new CairoEngine(configuration)) {
...
+try (CairoEngine engine = new CairoEngine(configuration)) {
+    final SqlExecutionContext ctx = new SqlExecutionContextImpl(engine, 1);
+    try (SqlCompiler compiler = new SqlCompiler(engine)) {
+        compiler.compile("create table testTable (" +
+                "a int, b byte, c short, d long, e float, g double, h date, " +
+                "i symbol, j string, k boolean, l geohash(8c), ts timestamp" +
+                ") timestamp(ts) partition by day WAL", ctx);
+
+        // write data into the table
+        try (TableWriterAPI writer = engine.getTableWriterAPI(ctx.getCairoSecurityContext(), "testTable")) {
```
glasstiger, line 213: this should be:
```
try (TableWriterAPI writer = engine.getTableWriterAPI(ctx.getCairoSecurityContext(), "testTable", "test")) {
```

docs/reference/api/java-embedded.md
glasstiger: correct, this is about the CREATE TABLE statement

docs/reference/api/java-embedded.md
```diff
@@ -164,43 +274,64 @@ always be 1.
...
+```java title="New table writer instance for a non-WAL table"
+
 try (TableWriter writer = engine.getWriter(ctx.getCairoSecurityContext(), "abc", "testing")) {
 ```
 
-The writer will hold exclusive lock on table `abc` until it is closed and
-`testing` will be used as the lock reason. This lock is both intra and
+```java title="New table writer instance for a WAL table"
+
+try (WalWriter writer = engine.getWriter(ctx.getCairoSecurityContext(), "abc", "testing")) {
```
glasstiger, line 309: should be:
```
try (WalWriter writer = engine.getWriter(ctx.getCairoSecurityContext(), "abc")) {
```

docs/reference/api/java-embedded.md
```diff
@@ -164,43 +274,64 @@ always be 1.
...
+```
+
+```java title="New table writer instance for either a WAL or non-WAL table"
+
+try (TableWriterAPI writer = engine.getWriter(ctx.getCairoSecurityContext(), "abc", "testing")) {
```
glasstiger, line 314: should be:
```
try (TableWriterAPI writer = engine.getTableWriterAPI(ctx.getCairoSecurityContext(), "abc", "testing")) {
```

docs/reference/api/java-embedded.md
glasstiger: yes.
WalWriter does not lock the table for write, concurrent data ingestion, modification and schema changes are allowed.

docs/reference/api/java-embedded.md
glasstiger: this is the same for all.
We should just mention it that regardless of the type of the writer new rows created the same way.

docs/reference/configuration.md
```diff
@@ -336,6 +336,8 @@ QuestDB.
 | cairo.sql.column.purge.retry.delay.limit       | 60000000          | Delay limit (μs), upon reaching which, the re-try delay remains constant.                                                                                                                                                |
 | cairo.sql.column.purge.retry.limit.days        | 31                | Number of days purge system will continue to re-try deleting stale column files before giving up.                                                                                                                        |
 | cairo.system.table.prefix                      | sys.              | Prefix of the tables used for QuestDB internal data storage. These tables are hidden from QuestDB web console.                                                                                                           |
+| cairo.wal.supported                            | false             | Setting enabling WAL table components for the system.                                                                                                             |
```
glasstiger, line 339: Setting enables the WAL (write ahead log) feature.
or:
Setting enables WAL-related components of the system.

docs/reference/api/java-embedded.md
amyshwang: So `WalWriter` does not "testing" but the other two do?

docs/concept/geohashes.md
```diff
@@ -318,6 +345,22 @@ try (TableWriter writer = engine.getWriter(ctx.getCairoSecurityContext(), "geoha
 
 ```
 
+```java title="TableWriterAPI"
+//Insert table into either a WAL or a non-WAL table:
+try (TableWriterAPI writer = engine.getWriter(ctx.getCairoSecurityContext(), "testTable", "test")) {
```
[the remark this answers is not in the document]
amyshwang: I realised that the table name for this section should be the same - geohash_table. 

I changed this code to:
```
try (TableWriterAPI writer = engine.getTableWriterAPI(ctx.getCairoSecurityContext(), "geohash_table")) {
``` 

And use the same table name for WalWriter - please check!

Also, TableWriterAPI does not need the `ApplyWal2TableJob` code?

docs/concept/write-ahead-log.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Key components
+
+A WAL table uses the following components to manage concurrent commit requests:
+
+- `WAL`: dedicated APIs for each ingestion interface. When data is ingested via
+  multiple interfaces, dedicated `WALs` ensure that the table is not locked for
+  one interface only.
+
+- `sequencer`: centrally manages transactions, providing a single source of
```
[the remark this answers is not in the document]
amyshwang: As this is not really used in any code (as opposed to `TableWriter`, I will remove the monospace. Same for WAL apply job!

docs/reference/api/java-embedded.md
glasstiger: yes, that is a lock reason, needed only for TableWriter.
WalWriter does not lock.

docs/concept/geohashes.md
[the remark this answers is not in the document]
glasstiger: TableWriter and TableWriterAPI needs the 3rd parameter (lock reason).
```
try (TableWriter writer = engine.getTableWriter(ctx.getCairoSecurityContext(), "geohash_table", "test")) {
```
```
try (WalWriter writer = engine.getWalWriter(ctx.getCairoSecurityContext(), "geohash_table")) {
```
```
try (TableWriterAPI writer = engine.getTableWriterAPI(ctx.getCairoSecurityContext(), "geohash_table", "test")) {
```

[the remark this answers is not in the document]
glasstiger: ...and let's add the apply job to both, WalWriter and TableWriterAPI.

docs/concept/write-ahead-log.md
```diff
@@ -0,0 +1,141 @@
...
+  - `ADD COLUMN` can only add 1 column per statement
+  - Non-structural operations may fail silently. These are partition-level and
+    configuration operations:
+
+    - `ATTACH PARTITION`
+    - `DETACH PARTITION`
+    - `DROP PARTITION`
+    - `SET PARAM`
+
+<!--TBD detach and attach-->
```
amyshwang, line 90:
```suggestion
```
Confirmed no further impact.

docs/concept/write-ahead-log.md
```diff
@@ -0,0 +1,140 @@
...
+
+- Sequencer: centrally manages transactions, providing a single source of truth.
+  The sequencer generates unique `txn` numbers as identifications to
+  transactions and keeps a log tracking the allocation. This log is called
+  `TransactionLog` and is stored in a meta file called `_txnlog`. See
+  [root directory](/docs/concept/root-directory-structure#db-directory) for more
+  information.
+
+- WAL apply job: collects the commit requests based on the unique `txn` numbers
+  and send them to the `TableWriter` to be committed.
```
puzpuzpuz, line 107:
```suggestion
  and sends them to the `TableWriter` to be committed.
```

docs/reference/configuration.md
```diff
@@ -337,6 +337,26 @@ QuestDB.
...
+### WAL table configurations
+
+When `cairo.wal.supported` is enabled, the following WAL tables settings on
+parallel threads are configurable for applying WAL data to the table storage:
+
+| Property                 | Default                        | Description                                                                                                                                                                                          |
+| ------------------------ | ------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| wal.apply.worker.count   | equal to the CPU core count    | Number of dedicated worker threads assigned to handle WAL tabel data. |
+| wal.apply.worker.affinity | equal to the CPU core count    | Comma separated list of CPU core indexes.                                            |
+| wal.apply.worker.yield.threshold | 10                    | TBD
```
ideoma, line 352: You can remove these 2 TBD lines. They are common for worker pool and we don't document them here, so let's not create precedence. 
Can you also add couple more from here 
https://github.com/questdb/questdb/blob/3ebdc1a8d87ceb436df686406ebccfede5d8178a/core/src/main/resources/io/questdb/site/conf/server.conf#L637

ideoma (dismissed):

docs/concept/write-ahead-log.md
```diff
@@ -0,0 +1,144 @@
...
+- Data ingestion through different interfaces
+- Data modifications
+- Table schema changes
+
+The following configurations and keywords enable and create WAL tables:
+
+<!--`cairo.wal.supported` will be removed.-->
+
+- WAL is an optional feature that is disabled by default. WAL can be enabled by
+  setting `cairo.wal.supported` to `true`.
```
glasstiger, line 29: this is `true` by default in 7.0.0 so it is already enabled

ideoma (approved):

glasstiger (approved):

PR #1225, opened 2022-12-07
chore(docs) DATRON refactoring

amyshwang:


src/pages/case-study/datron.tsx
```diff
@@ -141,6 +127,18 @@ const DATRON = () => {
             seCss["section--column"],
           )}
         >
+          <h3>About the team</h3>
```
amyshwang, line 130: I added this as a paragraph but I failed to delete the section with the same content:
(I broke the CI with my attempt):
image

PR #1231, opened 2022-12-12
chore(docs) document a new configuration option for Kafka connector

jerrinot:


docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -245,6 +246,27 @@ QuestDB supports a special type called
 configuration option to specify which columns should be created as the `symbol`
 type.
 
+### Numeric types inference
```
amyshwang, line 249:
```suggestion
### Numeric type inference
```

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -245,6 +246,27 @@ QuestDB supports a special type called
...
+
+When a configured Kafka Connect deserializer provides a schema then the connector will use it to determine column types. 
+If a schema is unavailable then the connector will infer the type from the value. This might produce unexpected
+results for floating point numbers. Consider this example:
```
amyshwang, line 253: Is this only going to be a problem for floating-point type?

amyshwang (commented):
Asking a question so that my suggestion will be on the right track!

docs/third-party-tools/kafka/questdb-kafka.md
jerrinot: yes, I think so. the problem is when you have floating point data, but the first observed sample happens not to have anything after a decimal point AND at the same time there is no explicit schema associated.  Then Kafka Connect creates a `long` object and QuestDB propagates it as a `long` column. 

The reverse case is OK: if the first sample is a floating point number then QuestDB will create a `double` column. When a next sample as a `long` value then it's OK, because QuestDB ILP receiver will do implicit casting from `long` to `double`.

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -185,6 +185,7 @@ The connector supports the following configuration options:
 | timestamp.units                   | `string`  | micros                                                      | auto               | Designated timestamp field units                           |
 | include.key                       | `boolean` | false                                                       | true               | Include message key in target table                        |
 | symbols                           | `string`  | instrument,stock                                            | N/A                | Comma separated list of columns that should be symbol type |
+| doubles                           | `string`  | volume,price                                                | N/A                | Comma separated list of columns that should be double type |
```
amyshwang, line 188:
```suggestion
| doubles                           | `string`  | volume, price                                                | N/A                | Comma separated list of columns that should be double type |
```

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -185,6 +185,7 @@ The connector supports the following configuration options:
 | timestamp.units                   | `string`  | micros                                                      | auto               | Designated timestamp field units                           |
 | include.key                       | `boolean` | false                                                       | true               | Include message key in target table                        |
 | symbols                           | `string`  | instrument,stock                                            | N/A                | Comma separated list of columns that should be symbol type |
```
amyshwang, line 187:
```suggestion
| symbols                           | `string`  | instrument, stock                                            | N/A                | Comma separated list of columns that should be symbol type |
```

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -245,6 +246,27 @@ QuestDB supports a special type called
 configuration option to specify which columns should be created as the `symbol`
 type.
 
+### Numeric type inference
```
amyshwang, line 249:
```suggestion
### Numeric type inference for floating point type
```

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -245,6 +246,27 @@ QuestDB supports a special type called
...
+
+When a configured Kafka Connect deserializer provides a schema then the connector will use it to determine column types. 
+If a schema is unavailable then the connector will infer the type from the value. This might produce unexpected
+results for floating point numbers. Consider this example:
```
amyshwang, line 253:
```suggestion
When a configured Kafka Connect deserializer provides a schema, the connector uses it to determine column types. 
If a schema is unavailable, the connector infers the type from the value. 
This might produce unexpected results for floating point numbers, which may be interpreted as `long` initially and generates an error. 

Consider this example:
```

amyshwang: I am trying to provide general rules before jumping into an example, although using an example is much easier...

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -245,6 +246,27 @@ QuestDB supports a special type called
...
+```json
+{ 
+  "instrument": "BTC-USD",
+  "volume": 42
+}
+```
+Kafka Connect JSON converter will deserialize the `volume` field as a `long` value. The connector will send it to
```
amyshwang, line 260:
```suggestion
Kafka Connect JSON converter deserializes the `volume` field as a `long` value. The connector sends it to
```
Previously we say "Kafka Connect deserializer" --> is the convertor part of the deserializer?

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -245,6 +246,27 @@ QuestDB supports a special type called
...
+QuestDB server as a `long` value. If the target table does not have a column `volume` then the database will create
+it as `long`. If the next message contains a floating point value for the `volume` field then the connector will send it
+to QuestDB as a `double` value. This will cause an error because the column `volume` already exists and it is of type
+`long`.
```
amyshwang, line 264:
```suggestion
QuestDB server as a `long` value. If the target table does not have a column `volume`, the database creates
a `long` column. If the next message contains a floating point value for the `volume` field, the connector sends it
to QuestDB as a `double` value. This causes an error because the existing column `volume`  is of type `long`.
```
- I'd avoid using future tense for tech documentation. If you're okay with it I can push a commit fix the rest of the page
- The last sentence is changed so that we don't give the impression that "the column already exists" may cause an error - I hope I understand this correctly.

docs/third-party-tools/kafka/questdb-kafka.md
```diff
@@ -245,6 +246,27 @@ QuestDB supports a special type called
...
+
+To avoid this problem, the connector can be configured to send selected numeric columns as `double` regardless of the
+value type. Use the `doubles` configuration option to specify which columns should the connector always send as the
```
amyshwang, line 267:
```suggestion
To avoid this problem, the connector can be configured to send selected numeric columns as `double` regardless of the actual initial input value. Use the `doubles` configuration option to specify which columns should the connector always send as the
```

Question: if the column should be of floating point type, the input value in the example above should be `42.00` to start with? Is it really a good idea to have a column with the values [42, 44.01, 40, 45.22], for instance?

amyshwang (changes requested):
Sorry I ask more questions......

docs/third-party-tools/kafka/questdb-kafka.md
jerrinot: hm, I would probably not do this. it's a configuration option. 
spaces in configuration options are dangerous - because different configuration management tools might interpret them in various ways. the connector itself won't mind it, because it trims whitespaces anyway, but I see no benefit in doing that.

docs/third-party-tools/kafka/questdb-kafka.md
jerrinot: yes, ideally, floating numbers should include `.00` or at least `.0` when they have no fraction part. this would make it unambiguous. 

unfortunately, this is not under our control. QuestDB just consumes from Kafka, and this is driven by whoever is producing the messages.

docs/third-party-tools/kafka/questdb-kafka.md
jerrinot: the same as the `symbols` option.

amyshwang (approved):

PR #1235, opened 2022-12-16
update privacy notice

argshook:
* remove excessive sections under Cloud policy of services that aren't used
* remove excessive items under Cloud collected data. That data was not actually collected, nor is intended to be collected in future.
* use correct `cloud.questdb.com` URL instead of `www.cloud.questdb.com`
* other stylistic changes to make document more readable
* typo fixes


amyshwang (approved):
Thank you!

PR #1237, opened 2022-12-17
improved docs comma punctuation

stletoss:
noticed this when reading through the data retention docs. Changed all three occasions of the same wording.


amyshwang (approved):
Thank you for your contribution!

PR #1245, opened 2022-12-30
feat(sql): added documentation for trigonometric functions sin, cos, tan, cot, asin, acos, atan, atan2, radians, degrees, pi

marregui:
Documents https://github.com/questdb/questdb/pull/2890


docs/reference/function/trigonometric.md
```diff
@@ -0,0 +1,300 @@
...
+## sin
+
+`sin(angleRadians)` returns the trigonometric sine of an angle. Special cases:
+
+- If the argument is NaN or an infinity, then the result is NaN.
+- If the argument is zero, then the result is a zero with the same sign as the argument.
+
+**Arguments:**
+
+- `angleRadians` numeric value for the angle, in radians.
```
amyshwang, line 19:
```suggestion
- `angleRadians` is a numeric value for the angle in radians.
```

docs/reference/function/trigonometric.md
```diff
@@ -0,0 +1,300 @@
...
+
+**Return value:**
+
+The sine of the argument, of type `Double``.
```
amyshwang, line 23:
```suggestion
Return value type is `double`.
```

docs/reference/function/trigonometric.md
```diff
@@ -0,0 +1,300 @@
...
+
+This page describes the available functions to assist with performing trigonometric
+calculations.
+
+## sin
+
+`sin(angleRadians)` returns the trigonometric sine of an angle. Special cases:
+
+- If the argument is NaN or an infinity, then the result is NaN.
+- If the argument is zero, then the result is a zero with the same sign as the argument.
```
amyshwang, line 15: I am not sure if `sin 0 = 0` is a special case. I don't know what you mean "with the same sign as the argument" though.

docs/reference/function/trigonometric.md
```diff
@@ -0,0 +1,300 @@
...
+| angle              | sin |
+|--------------------|-----|
+| 1.5707963267948966 | 1.0 |
+
+## cos
+
+`cos(angleRadians)` returns the trigonometric cosine of an angle. Special cases:
+
+- If the argument is NaN or an infinity, then the result is NaN.
+- If the argument is zero, then the result is a 1.0.
```
amyshwang, line 40: Again this is always the case not specific to this function - do we need to mention this?

docs/reference/function/trigonometric.md
```diff
@@ -0,0 +1,300 @@
...
+
+**Examples:**
+
+```questdb-sql
+SELECT atan(1.0) atan;
+```
+
+| atan               |
+|--------------------|
+| 0.7853981633974483 |
```
amyshwang, line 234: This does not make sense to me: there should be 2 inputs and the function is `tan2`?

docs/reference/function/trigonometric.md
```diff
@@ -0,0 +1,300 @@
...
+
+## atan2
+
+`atan2(valueY, valueX)` returns the angle *theta* from the conversion of rectangular coordinates (x, y) to polar (r, theta).
```
amyshwang, line 188: The input is asking for (y, x) so the description might need to change to:
```suggestion
`atan2(valueY, valueX)` returns the angle *theta* from the conversion of rectangular coordinates (y, x) to polar (r, theta).
```
I am also not sure the design for the input to be (y, x) - (x, y) is more intuitive?

docs/reference/function/trigonometric.md
```diff
@@ -0,0 +1,300 @@
...
+- `angleRadians` numeric value for the angle, in radians.
+
+**Return value:**
+
+The sine of the argument, of type `Double``.
+
+**Examples:**
+
+```questdb-sql
+SELECT PI()/2 angle, sin(PI()/2) sin;
```
amyshwang, line 28: PI should be pi?

amyshwang (changes requested):
I will add commits for general formatting and wording but some questions to be clarified:

1. Some special cases are just how trigonometric works - do we need to specify them? (e.g. sin 0 = 0)
2. `atan2` has wrong example. Also I found the input (y, x) strange...

docs/reference/function/trigonometric.md
amyshwang: Checked PostgreSQL atan2 and this might work better:
```
Inverse tangent of y/x, result in radians
```

docs/reference/function/trigonometric.md
marregui:
```sql
select sin(0.0), sin(-0.0), cos(0.0), cos(-0.0), tan(0.0), tan(-0.0)
```
| sin | sin1  | cos | cos1 | tan | tan1 |
|----|-----|-----|------|----|------|
| 0.0 | -0.0 | 1.0  | 1.0.   | 0.0 | -0.0 |

docs/reference/function/trigonometric.md
marregui: this documents the expected result for those cases

docs/reference/function/trigonometric.md
marregui: lol. this is just a pass through call to Math.atan2, with documentation extracted verbatim.

docs/reference/function/trigonometric.md
marregui: ok

docs/reference/function/trigonometric.md
amyshwang: `cos0 = 1` should not need special documentation as this is just math? Unless you meant that it is special in that it is a double, `1.0` instead of `1` - in which case the "return value" section conveys it already?

docs/reference/function/trigonometric.md
```diff
@@ -0,0 +1,318 @@
...
+`atan2(valueY, valueX)` returns the angle _theta_ from the conversion of
+rectangular coordinates (x, y) to polar (r, theta). This function computes
+_theta_ (the phase) by computing an arctangent of y/x in the range of -pi to pi.
+Special cases:
+
+- If either argument is `NaN`, then the result is `NaN`.
+- If the first argument is positive zero and the second argument is positive, or
+  the first argument is positive and finite and the second argument is positive
+  infinity, then the result is positive zero.
+- If the first argument is negative zero and the second argument is positive, or
```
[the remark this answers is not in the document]
marregui: bug in web console

amyshwang (approved):

PR #1247, opened 2023-01-04
chore(docs): adding FAQ for HikariCP config

amyshwang:


amyshwang:
As discussed, one of our users contributes the problem to the timeout setting but after some investigation, we need to confirm the actual root cause before adding this to FAQ. We have decided to close this PR.

PR #1248, opened 2023-01-04
chore(docs) improving the IN operator wording

### LINKED ISSUE #1234: Timestamp in: Time range with modifier unclear
I'm reading https://questdb.io/docs/reference/sql/where/#time-range-with-modifier and I'm not getting it. 

There are multiple issues: 
1. This is not clear:
  > You can apply a modifier to further customize the range. The algorithm will calculate the resulting range by modifying the upper bound of the original range by the modifier parameter. An optional occurrence can be set to apply the time range repeatedly for a set number of times.
  I have a hard time to parse. Perhaps there should be an explanation of what it is good for. That would help me to understand it.
2. The legend does not match the grammar graph:

  image

  What is "period"? `Period` is not in the grammar graph at all. 

3. Each item of the legend should have a short description. Right now it just says what is a type of each parameter. But it does not explain what's the meaning of the parameters. 
4. This example is rather complicated and deserves more details explanation. I don't get the meaning of each parameter in `'2018-01-01;1d;1y;2';`
  image

amyshwang:
Closes #1234


docs/reference/sql/where.md
```diff
@@ -366,7 +371,9 @@ Modifying the interval:
 SELECT * FROM scores WHERE ts IN '2018-01-01;1d;1y;2';
 
 ```
-The range is Jan 1 2018 with a one-year interval and the total occurrence is two.
+
+The range, Jan 1 2018 (one day), with a one-year interval, repeated twice. This
```
jerrinot, line 375: I think there is a mistake here even in the original text. 
'2018-01-01;1d;1y;2'; indeed sets the date to `Jan 1 2018`, one-year interval, repeated twice. But it also sets the modified to 1 day. so:
1. this text should explicitly say there is a modified 1 day
2. it should be reflected in the table - where the 1st interval should end with `2018-01-02T23:59:59.999999Z` and the 2nd interval with `2019-01-02T23:59:59.999999Z`. Unless I am missing something terribly. 

The alternative is to keep the table as it is and set the modified to `0d`.

jerrinot (changes requested):
this PR greatly improves clarity! 🎉 

in fact it improves it so much that I finally understood how it works, and this helped me to find a mistake in one of the examples! :)

docs/reference/sql/where.md
jerrinot: maybe something like: "the modifier set to `1d` extends the upper bound of each interval by 1 day".

jerrinot (approved):

PR #1280, opened 2023-01-10
chore(docs): added explain description

bziobrowski:


amyshwang:
Notes from my commit:

1.
```
- `Filter` - standalone (non-JIT-ted, non-parallelized) filter.
```
Is `non-JIT-ted` intentional or is it a typo?

2.
```
- `Limit` - standalone node implementing the `LIMIT` keyword. Other nodes can implement LIMIT` internally, e.g. Sort 
```
I gather `Sort` here is not an SQL keyword. It is one of the node types. However in the main page the capitalisation is not consistent, sometimes it is in monospace too. I get confused about which sort means the `Sort` node. We could maybe search this word in the file and:
- For the node, use one of the following:
```
the `sort` node
the `Sort` node
```
or
```
the `sort` buffer
the `Sort` buffer
```
- For other sort meaning sorting data, just normal format.

It is possible that all the "sorts" in the file refer to the node! I don't have a preference for `sort` and `Sort` but I noted inconsistencies.

3. 
```
## Limitations:

To minimize resource usage, the `EXPLAIN` command does not execute the statement,
to avoid paying a potentially large upfront cost for certain queries
(especially those involving hash, join. or sort) and making the plan less
accurate.  
```

My rewriting assumes that EXPLAIN does not execute the statement in order to avoid two things:
- paying a large cost
- making it (EXPLAIN itself, therefore the plan) less accurate

I want to check my understanding is correct...

amyshwang:
commit [491c92e](https://github.com/questdb/questdb.io/pull/1280/commits/491c92ee0c3055dcebac0380a03f83c8ad0088ef) tries to fix the missing elements in the diagram by copying defs from another svg file. I did not check what went wrong exactly but after re-copying the first part of the SVG file, it seems to fix it:

Original:

image

Local build for the change:

image

bziobrowski:
> Notes from my commit:
> 
> 
> ```
> - `Filter` - standalone (non-JIT-ted, non-parallelized) filter.
> ```
> 
> Is `non-JIT-ted` intentional or is it a typo?

No, it's intentional . Probably too colloquial . How using `JIT-compiled/non-JIT-compiled` everywhere ? 

> 
> ```
> - `Limit` - standalone node implementing the `LIMIT` keyword. Other nodes can implement LIMIT` internally, e.g. Sort 
> ```
> 
> I gather `Sort` here is not an SQL keyword. It is one of the node types. However in the main page the capitalisation is not consistent, sometimes it is in monospace too. I get confused about which sort means the `Sort` node. We could maybe search this word in the file and:
> 
> * For the node, use one of the following:
> 
> ```
> the `sort` node
> the `Sort` node
> ```
> 
> or
> 
> ```
> the `sort` buffer
> the `Sort` buffer
> ```
> 
> * For other sort meaning sorting data, just normal format.
> 

Yes, sort here means `Sort' execution plan node . I think other usages of 'sort'  are fine because they don't mean sort node but sorting operation in general . 

> It is possible that all the "sorts" in the file refer to the node! I don't have a preference for `sort` and `Sort` but I noted inconsistencies.
> 
> 
> ```
> ## Limitations:
> 
> To minimize resource usage, the `EXPLAIN` command does not execute the statement,
> to avoid paying a potentially large upfront cost for certain queries
> (especially those involving hash, join. or sort) and making the plan less
> accurate.  
> ```
> 
> My rewriting assumes that EXPLAIN does not execute the statement in order to avoid two things:
> 
> * paying a large cost
> * making it (EXPLAIN itself, therefore the plan) less accurate
> 
> I want to check my understanding is correct...


It's only to avoid executing the query, which could take a long time (time out even). 
Because explain doesn't create and use cursors it's missing things that are determined at query runtime and that reduces accuracy .
Actual execution plan of executed query could be slightly different from one shown by explain but that's fine because many databases are the same, e.g. PostgreSQL  .  Postgres has a variant, EXPLAIN ANALYZE, to run the query with instrumentation to produce exact plan with runtime stats .

docs/reference/sql/explain.md
```diff
@@ -0,0 +1,252 @@
...
+In this example, the plan shows that the `trades` table undergoes a full scan
+(`DataFrame` and subnodes) and the data is processed by the parallelized JIT-ted
+filter.
+
+### Using `EXPLAIN` for the plan for `CREATE` and `INSERT`
+
+Apart from `SELECT`, `EXPLAIN` also works on `CREATE` and `INSERT` statements.
+Single-row inserts are straightforward. The examples in this section show the
+plan for more complicated `CREATE` and `INSERT` queries.
+
```
bziobrowski, line 158: @amyshwang  Do we need to show tab content here ?

amyshwang: I added it because we move away from the previous table. But it is silly. I will just add some wording...

amyshwang:
The last commit reworded Limitation, the use of "sort" and some minor editorial changes.

amyshwang (approved):

PR #1296, opened 2023-01-12
chore(ui): update blog post page html structure

argshook:
This PR tweaks the blog post page HTML which is used to render pages like https://questdb.io/blog/2022/11/08/questdb-devoxx-belgium-2022

There are visible and invisible changes.

First, visible changes

# Visible changes

* (most notable change) move `Recent Posts` sidebar to the right side of the page
* since website is written in left-to-right language (English) it makes sense to start the content on the left side.
* expand width of main article slightly, to be same width as it is on `/docs`
* add `edit this page` link at the bottom which navigates to relevant github edit page

the alignment of content is now more similar to that on `/docs`. Thus, switching between pages is more visually consistent

# Screenshots

## before:

Screen Shot 2023-01-13 at 00 44 18

## after:

Screen Shot 2023-01-13 at 00 41 41

## `Edit this page` link

* easier for maintainers to quickly submit a fix
* might encourage non-maintainer visitors to contribute

# Invisible changes

The HTML is tweaked in the following ways:

* make main content appear first in HTML
  * previously `Recent articles` would appear first in HTML, now the main content goes first. This is also a small nudge for SEO.
* reduce number of nested divs for layout. Solve that, where possible, with CSS grid
* use HTML5 elements like `main`, `aside`, `header`, `footer` where appropriate (instead of `div`)
* use own CSS instead of globally available one, to make it easier in future to remove unused CSS


amyshwang:
I think this is a good idea!
Looking at the preview page however, I wonder if we should show the sections of the article, like we do in Docs:
image

Following the same logic, on the left-hand side, we could keep recent articles. On the right-hand side, a structure of the article currently viewed?

insmac (approved):

argshook:
> I think this is a good idea!
> Looking at the preview page however, I wonder if we should show the sections of the article, like we do in Docs:
> image
> 
> Following the same logic, on the left-hand side, we could keep recent articles. On the right-hand side, a structure of the article currently viewed?

Hey @amyshwang, thanks for checking the PR!

Your idea is good and I tried to implement it while working on this PR.

One thing I dislike is the narrowing of content area with additional sidebar, which is not as useful on blog posts as it is on documentation.

Blog post is usually read from top to bottom as a whole, while documentation might also be read in a reference kind of way, by shifting focus to multiple areas. In there, displaying the outline of the document is very useful.

In addition, blog posts usually have just a few headers, whilst docs might have multiple and even multilevel ones.

On the end I came to this conclusion: displaying outline on blog posts could still be useful but it's not useful enough to dedicate a whole sidebar for it. This, it could be displayed in the same sidebar where `Recent posts` are.

Unfortunately this is not a trivial thing to implement, need to override some docusaurus things. I'd be happy to do that in a separate PR.

Obligatory: sorry for wall of text :D

amyshwang:
Thanks for your explanation - it's long but really useful. I think you have a good point there. I am on a super big screen so the left-hand side looks very empty. But on a normal screen, we probably should leave one side free of content! :)

argshook:
> I am on a super big screen so the left-hand side looks very empty.

Side question: on wide screens the left sidebar on `/docs` is aaall the way attached to the left side of screen. I always thought it was a bug, but it's actually the intended way by docusaurus.

I would move that sidebar to be closer to content, what would you think about such UI?

PR #1299, opened 2023-01-13
chore(docs) Adding IN to FAQ

amyshwang:


docs/troubleshooting/faq.md
```diff
@@ -88,3 +88,19 @@ We have an open
 inserted with identical fields. Until then, you need to
 [modify the data](/docs/guides/modifying-data) after it's inserted and use a
 `GROUP BY` query to identify duplicates.
+
+## How do I query recursively within a time range?
```
jerrinot, line 92: nit: is it possible to use simpler language for the title? the question we keep getting is this: "Can we query by time?" "querying recursively within a time range" is a rather complex sentence to parse.

jerrinot (dismissed):
I'd personally use simpler language, but that could be just a personal taste.

docs/troubleshooting/faq.md
amyshwang, line 92:
```suggestion
## Can I query by time?
```

docs/troubleshooting/faq.md
```diff
@@ -88,3 +88,19 @@ We have an open
...
+
+When using the `WHERE` statement to define the time range for a query, the
```
amyshwang, line 94:
```suggestion
Yes! When using the `WHERE` statement to define the time range for a query, the
```

docs/troubleshooting/faq.md
amyshwang: Good point - updated!

jerrinot (approved):
now it's pretty awesome: simple and straight to the point. 
thank you!

PR #1301, opened 2023-01-13
chore(docs) `cairo.o3.min.lag` config

amyshwang:
1. in `server.conf` I cannot find the min and max setting
2. I found the default values [here](https://github.com/questdb/questdb/blob/5d36b8348cb61ff1c38734a8ffd26df0e37656bc/core/src/main/java/io/questdb/cairo/DefaultCairoConfiguration.java#L355-L364)

```
  @Override
    public long getO3MaxLag() {
        // 5 min
        return 300_000_000L;
    }

    @Override
    public long getO3MinLag() {
        return 1_000_000;
    }
```
I assume the min default is 1000000 milliseconds.
For max, we have 10 minutes in the docs but here it says 5 mins in commentary but the value `300_000_000` is 50000 mins?


puzpuzpuz (approved):
Thanks!

PR #1306, opened 2023-01-16
feat(docs) uuid type documented

jerrinot:


amyshwang:
Question for @jerrinot : is there a `NULL` value to be added to the table?

jerrinot:
@amyshwang good one. added. thanks!

amyshwang (approved):

PR #1313, opened 2023-01-16
Use inline styles for home subheader

insmac:
Performance enhancement: adjusting the largest contentful paint to use inline styles.


insmac:
Closing, as this does not yield any measurable performance improvement. Furthermore, it does not support media queries, which is a no-go.

PR #1331, opened 2023-01-23
chore(docs): update rolling log writer config sample

puzpuzpuz:
A follow-up for https://github.com/questdb/questdb/pull/2740


docs/reference/configuration.md
```diff
@@ -560,7 +560,12 @@ writers=file,stdout
 #w.file.level=INFO,ERROR
 #rollEvery accepts: day, hour, minute, month
 #w.file.rollEvery=day
-#w.file.rollSize=1g
+#rollSize specifies size at which to roll a new log file: a number followed by k, m, g (KB, MB, GB respectively)
```
amyshwang, line 563: Not a change request but I wonder if it is common to add a comment/explanation before the setting/configuration? 
I am pointing this out because I intuitively thought line 563 explains line 562.
It is probably just me. There are no other examples on this page.

amyshwang (approved):

docs/reference/configuration.md
puzpuzpuz: Yes, that's what we have in the default `log.conf` file that gets generated on the first database start.

PR #1334, opened 2023-01-23
chore(docs) hiding eq() which we don't support

amyshwang:


docs/reference/sql/where.md
```diff
@@ -165,6 +165,7 @@ SELECT * FROM users WHERE age = 23;
 SELECT * FROM users WHERE age != 23;
 ```
 
+<!--
```
puzpuzpuz, line 168: Would be nice to leave a clarifying comment here. If we keep this section commented, the comment should describe why it happened and under which circumstances we hope to uncomment it in future.

amyshwang:
```suggestion
<!-- 
QuestDB does not support `eq()`. This section is therefore commented out and can be uncommented when we add the functionality.
```

puzpuzpuz (approved):

PR #1341, opened 2023-01-25
feat(docs) adding CREATE TABLE IN VOLUME docs

amyshwang:
Documentation for https://github.com/questdb/questdb/pull/2931


docs/reference/sql/create-table.md
```diff
@@ -247,7 +242,60 @@ SELECT id, name, maxUncommittedRows FROM tables();
...
+  ```questdb-sql
+  CREATE TABLE my_table (i symbol, ts timestamp), index(i capacity 32) WITH maxUncommittedRows=7, IN VOLUME SECONDARY_VOLUME;
+  ```
+
+- If no `WITH` clause is used, the comma must not be added for the `IN VOLUME`
+  segment:
+
+  ```questdb-sql
+  CREATE TABLE my_table (i symbol, ts timestamp) IN VOLUME SECONDARY_VOLUME;
+  ```
```
amyshwang, line 267: I think this is also okay, based on the code PR:

```suggestion
  ```questdb-sql
  CREATE TABLE my_table (i symbol, ts timestamp) IN VOLUME 'SECONDARY_VOLUME';
  ```
```
However, I think it is a bit over-complicated so I am not adding this to the syntax diagram. Could we leave it out from the docs?

marregui: ofcourse, I do not think i adds much in the diagrams. The gist of this is that if you define a volume:

```
other volume -> /absolute/path/to/other/volume
```

then you can:

```
CREATE TABLE my_table ( ... ) ... IN VOLUME 'other volume';
CREATE TABLE my_table ( ... ) ... IN VOLUME 'other voluMe';
...
```

If alias was just `othervolume`, or `other_volume`, then this would be possible:

```
CREATE TABLE my_table ( ... ) ... IN VOLUME other_volume;
CREATE TABLE my_table ( ... ) ... IN VOLUME othervolume;
...
```

I expect users to define aliases that make sense for them, omitting quotes is just nice (I forgot them all the time while testing).

marregui (dismissed):
LGTM

docs/reference/sql/drop.md
```diff
@@ -8,8 +8,12 @@ description: DROP TABLE SQL keyword reference documentation.
...
 :::caution
 
-This command irremediably deletes the data in the target table. In doubt, make
-sure you have created [backups](/docs/reference/sql/backup) of your data.
+This command irremediably deletes the data in the target table. Unless the 
+table was created in a different volume than the standard, see 
+[IN VOLUME](/docs/reference/sql/create-table/#table-target-volume), in
+which case the table is only logically removed and data remains intact in 
+its volume. In doubt, make sure you have created 
+[backups](/docs/reference/sql/backup) of your data.
```
bluestreak01, line 16: trailing slash missing

bluestreak01 (changes requested):

docs/reference/sql/drop.md
bluestreak01, line 16:
```suggestion
[backups](/docs/reference/sql/backup/) of your data.
```

bluestreak01 (dismissed):
missing trailing slash

docs/reference/sql/create-table.md
amyshwang: Ah in this case they are used in a specific situation so I'd add the option to the diagram.

amyshwang: Diagram and text updated!

marregui (approved):
I have added a few missing slashes in the md docs affected by this review, as my understanding is it helps.

glasstiger (approved):

PR #1343, opened 2023-01-25
chore(blog/tags/*): set specific page title for blog tags pages

argshook:
This PR addresses the problem of duplicate titles on tags pages.

It is not beneficial for SEO to have duplicate titles on pages. This PR addresses this issue by adding the tag name to the title of the page.

For example, this change will make the title of page like `/blog/tags/release` as following:

```diff
- QuestDB
+ Posts tagged with "Release" | QuestDB
```


bluestreak01:
good job @argshook also fixes trailing `/` !

bluestreak01 (approved):

PR #1356, opened 2023-01-27
fix(docs): fix root directory structure for Mac

insmac:
Current value for MacOS directory structure only works for ARM (M1/M2) Macs, Intel machines have a different path.


docs/concept/root-directory-structure.md
```diff
@@ -44,7 +44,8 @@ $HOME/.questdb
 
 
 ```shell
-/opt/homebrew/var/questdb
+M1/M2: /opt/homebrew/var/questdb
+Intel: /usr/local/var/questdb
```
argshook, line 48: I would change this a bit to display paths separately. When content is
shown in a code block, it might give an impression to the reader, that
it's part of some code or output.

But this is not the case here. Here's how i'd write such instances:

    Path on macbooks with M1 or M2 chip:
    ```shell
    /opt/homebrew/var/questdb
    ```
    
    Path on macbooks with M1 or M2 chip:
    ```shell
    /usr/local/var/questdb
    ```

Bit more verbose, but intent is clear.

(unable to post this as github suggestion)

argshook (changes requested):

argshook (approved):

PR #1383, opened 2023-02-07
WIP: Adds cloud journey post

sklarsa:


blog/2023-02-24-cloud-journey.mdx
```diff
@@ -0,0 +1,224 @@
...
+led us to some admittedly industry-standard choices in our technology decisions.
+
+### Architecture
+
+We decided to initially build our service on AWS for a simple reason; most of our prospective clients were also running there! Since
+we pride ourselves on ingestion and query speed, we didn't want to introduce additional uncontrollable variables like network
+latency that would not give customers the performance that they came to us for in the first place.
+
+This led us to design a multi-region framework, allowing customers to deploy their own databases physically close to their infrastructure.
+In order to do this effectively, we knew that we would need to employ IaC tools like Terraform to provision infrastructure
```
amyshwang, line 151: Maybe it is my lack of knowledge but I feel this is very vague. I had to google what IaC is, which is fine but I wonder if we could position this as a general knowledge article about cloud and introduce this a bit more?

amyshwang (commented):
I agree with your view that this article would be better off splitting into:
- Cloud UI
- Cloud (core?)
- Metrics and logging (already covered)
With the Cloud core section, you have provided what tools we use and that is all. I am curious about:
- what are some challenges in this design choice?
- what are some other tools we have considered?
Basically a bit more explanation for people who are technical but not in the cloud space and with a bit of a story, if this makes sense at all...

sklarsa:
@amyshwang I'm going to close this PR because I'm going to pivot the infrastructure article.  I'm going to craft a reply to [this tweet](https://twitter.com/kelseyhightower/status/1624096644865327105?t=toBgQADIX1S8QUEz1GDCsw) which was making the rounds last week about hosting DBs on k8s.  I think it would be nice to use that to create a narrative about why it's possible, but hard, to host DBs in k8s and what makes our cloud so great.  As a bonus, it could spark a flame war in the comments which is always good for visibility :-)

Issue #1348, opened 2023-01-25
InfluxDB to QuestDB


amyshwang:

amyshwang:
Create a page in Docs to capture - https://questdb.slab.com/posts/migration-influx-db-to-quest-db-o9fsp6g3

PR #1386, opened 2023-02-09
chore(docs)influxdb migration

amyshwang:
Part one for #1348


docs/guides/influxdb-migration.md
```diff
@@ -0,0 +1,157 @@
...
+several metrics for the same series, you might want to transform your data in
+QuestDB.
+
+For example, if you query a table with several metrics:
+
+```questdb-sql
+SELECT * FROM diagnostics WHERE timestamp = '2016-01-01T00:00:00.000000Z' AND driver='Andy' AND name='truck_150')
+```
+
+Your result may be something like this:
```
amyshwang, line 107: What is 'this'? Duplicate rows? It is not clear to me. I think we need to be more descriptive, in order to update the alt too!

amyshwang (commented):
@javier I migrated your notes to our doc but further info is required...

docs/guides/influxdb-migration.md
```diff
@@ -0,0 +1,157 @@
...
+FROM
+  diagnostics;
+```
+
+This produces an aggregated row containing all the metrics for each dimension
+group:
+
+![Screenshot of the query result](/img/docs/guide/adjusted-metric.png)
+
+You can use the `SELECT INTO` functionality in QuestDB to output the processed
```
amyshwang, line 137: Do you mean INSERT INTO?

docs/guides/influxdb-migration.md
```diff
@@ -0,0 +1,157 @@
...
+
+For example, if you query a table with several metrics:
+
+```questdb-sql
+SELECT * FROM diagnostics WHERE timestamp = '2016-01-01T00:00:00.000000Z' AND driver='Andy' AND name='truck_150')
+```
+
+Your result may be something like this:
+
+![Screenshot of the query result](/img/docs/guide/one-metric.png)
```
amyshwang, line 109: Screenshot needs cropping once I understand what to focus on

docs/guides/influxdb-migration.md
```diff
@@ -0,0 +1,157 @@
...
+  max(nominal_fuel_consumption) AS nominal_fuel_consumption,
+  max(status) AS status
+FROM
+  diagnostics;
+```
+
+This produces an aggregated row containing all the metrics for each dimension
+group:
+
+![Screenshot of the query result](/img/docs/guide/adjusted-metric.png)
```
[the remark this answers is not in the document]
amyshwang: Needs cropping as image is too small to read.

docs/guides/influxdb-migration.md
javier: No duplicates. The way influx exports the data as ILP is exporting a single metric on each line. So it exports all the dimensions, and then one metric. If your table has several metrics, then you will have for each logical row several individual ILP lines. When ingesting, you will have a line for each metric, with all the other metric values showing NULL. As an example, let's say I am exporting a table that would look like

Andy, truck_150, 50, 100, timestampX. 

Where 50 is speed and 100 is temperature

When you export in Influx you will get two lines

Andy, truck_150, 50, NULL, timestampX. 
Andy, truck_150, NULL, 100, timestampX. 

So basically you probably want to recompose into a single line, which is what my query at Slab would do. It aggregates all the metrics that happened for the same timestamp and the same dimension values, and just consolidates into a single row.

I hope this makes sense

docs/guides/influxdb-migration.md
javier: Yes. Thanks. In standard SQL you can either INSERT INTO SELECT, or just SELECT INTO. But I keep forgetting in QuestDB we only allow the former. Sorry

docs/guides/influxdb-migration.md
```diff
@@ -0,0 +1,157 @@
...
+
+You can use the `SELECT INTO` functionality in QuestDB to output the processed
+result into a new table.
+
+### Importing small tables
+
+If you have small tables to migrate, importing data can be easier by using
+InfluxDB to query the data and outputting the result to CSV. The CSV can be
+[imported into QuestDB](/docs/guides/importing-data-rest/). This doesn’t work
+for large tables.
```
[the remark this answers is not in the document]
javier: We do recommend it, but only for small tables. Large tables will give you issues both when exporting from Influx and when importing into QuestDB

docs/guides/influxdb-migration.md
amyshwang: Thank you! Content updated!

docs/guides/influxdb-migration.md
amyshwang: No probs! Updated.

sidebars.js
```diff
@@ -56,6 +56,7 @@ module.exports = {
       label: "Guides",
       type: "category",
       items: [
+        "guides/influxdb-migration",
         "guides/importing-data",
         "guides/importing-data-rest",
         "guides/modifying-data",
```
amyshwang, line 62:
```suggestion
        "guides/importing-data",
        "guides/importing-data-rest",
        "guides/influxdb-migration",
        "guides/modifying-data",
```

javier (approved):

PR #1388, opened 2023-02-09
chore(docs) document rnd_uuid4()

amyshwang:


jerrinot:
hi Amy, good catch, many thanks for starting this!

we should also specify:
1. the function generates UUID version 4
2. the generated UUID should not be used as a secret. 

I'll push an update to this PR.

PR #1397, opened 2023-02-16
feat(docs): redpanda integration guide

### LINKED ISSUE #1387: Update - red panda integration
**Describe the requested changes:**

Red panda integration page

Some thoughts about the page structure:

- What is Redpanda
- An example replacing a Kafka connector with redpanda - maybe using [the example of our kafka connector](https://github.com/questdb/kafka-questdb-connector/blob/main/kafka-questdb-connector-samples/stocks/docker-compose.yml)
- reference: https://docs.redpanda.com/docs/get-started/quick-start/quick-start-docker/
Question: is this possible to do at all?
- See also links
  - We can elaborate this idea into a separate blog post:
```
How to import Parquet files to QuestDB.
QuestDB does not support Parquet files directly, but I think it should be possible to import them via the Pandas support. As Parquet files can be transformed to Pandas (I think I saw it in one of the Adam’s examples). I think that’s currently the most efficient way to load Parquet files to QuestDB
```
I think the parquet example is better suited as a tutorial as using redpanda is not necessarily the best way of doing it; it's one of many methods.

amyshwang:
closes #1387


docs/third-party-tools/redpanda.md
```diff
@@ -0,0 +1,251 @@
...
+
+To stop the containers, run:
+
+```shell
+docker compose down
+```
+
+## See also
+
+- [How to build a real-time crypto tracker with Redpanda and QuestDB](https://redpanda.com/blog/real-time-crypto-tracker-questdb-redpanda)
```
jerrinot, line 251: consider linking the QuestDB Kafka Connector reference manual: https://deploy-preview-1397--unruffled-blackwell-31bfb2.netlify.app/docs/third-party-tools/kafka/questdb-kafka/#configuration-manual

docs/third-party-tools/redpanda.md
```diff
@@ -0,0 +1,250 @@
...
+The message is delivered to QuestDB:
+
+![QuestDB web console result showing the Redpanda message](/img/docs/guide/redpanda/questdb-select.png)
+
+## Terminate Docker
+
+To stop the containers, run:
+
+```shell
+docker compose down
```
jerrinot, line 244: is this really needed? you are starting docker compose with `docker compose up` (=the default attached mode). So killing this should be just something like ctrl+c? I believe `docker compose down` is used when you start it the detached mode `docker compose up -d`

amyshwang: Ah thanks it makes sense!

jerrinot (approved):
there is this `docker compose down` question. otherwise, it looks good. well done!

jerrinot (approved):

PR #1403, opened 2023-02-17
chore(docs): update python script to use psycopg3

amyshwang:


amunra (approved):
You've changed the example query to something simpler, and I agree with you: I think it works better and it's more representative. Also good that you're now using `with` statements which are indeed the proper approach.
Nice!

Issue #1396, opened 2023-02-16
Inconsistent terminology in the JOIN description


jerrinot:
LT JOIN and ASOF JOIN are very similar. This is what the page says about [ASOF joins](https://questdb.io/docs/reference/sql/join/#asof-join):
> ASOF joins are used on time series data to join two tables based on timestamps that do not exactly match. For a given record at a given timestamp, it will return the corresponding record in the **other** table at the closest timestamp prior to the timestamp in the **first** table.

and this is what it says about [LT joins](https://questdb.io/docs/reference/sql/join/#lt-join):
> LT join is very similar to ASOF, except that it searches for the last row from the **right** table strictly before the row from the **left** table. There will be one or no rows joined from the **right** table per each row from the **left** table.

So for the same concept, we use first/second and left/right. I find it confusing, we should be consistent.  
In fact, I had to think twice to realize which table is considered left and which is right, but this could be just my SQL ignorance.

jerrinot:
It looks there is a more serious problem. The ASOF JOIN description says this:
>  For a given record at a given timestamp, it will return the corresponding record in the other table at the closest timestamp **prior to** the timestamp in the first table.

It's not "prior to", is it? the it's "prior to or equal". At least that's my understand of the difference between ASOF and LT join:
1. ASOF = prior or equal
2. LT = strictly prior

Caveat: Chances are I got it wrong 🤷

PR #1406, opened 2023-02-20
chore(docs): improve LT and ASOF JOINS documentation

jerrinot:
fixes https://github.com/questdb/questdb.io/issues/1396


jerrinot:
This is still WIP, more changes to be done. opening a draft PR for visibility. so @amyshwang can tell me this is all rubbish :)))

docs/reference/sql/join.md
```diff
@@ -168,72 +171,71 @@ CROSS JOIN deserts;
...
 | ts                          | bid |
-| --------------------------- | --- |
+|-----------------------------|-----|
+| 2019-10-17T00:00:00.000000Z | 100 |
 | 2019-10-17T00:00:00.100000Z | 101 |
 | 2019-10-17T00:00:00.300000Z | 102 |
 | 2019-10-17T00:00:00.500000Z | 103 |
+| 2019-10-17T00:00:00.600000Z | 104 |
+
+Table `asks`:
```
jerrinot, line 197: I change the order of table: first the `bids` table and then the `asks` table. Why? that's the order used in the SQL: 
```sql
[...]
FROM bids
ASOF JOIN asks;
```
First bids and then asks.

docs/reference/sql/join.md
```diff
@@ -168,72 +171,71 @@ CROSS JOIN deserts;
...
+|-----------------------------|-----|
+| 2019-10-17T00:00:00.100000Z | 100 |
+| 2019-10-17T00:00:00.300000Z | 101 |
+| 2019-10-17T00:00:00.400000Z | 102 |
 
 An `ASOF JOIN` query can look like the following:
 
 ```questdb-sql
-SELECT bids.ts timebid, bid, ask
+SELECT bids.ts timebid, asks.ts timeask, bid, ask
```
jerrinot, line 208: I believe including the `timeask` column makes it easier to see how the join worked.

docs/reference/sql/join.md
```diff
@@ -168,72 +171,71 @@ CROSS JOIN deserts;
...
+| 2019-10-17T00:00:00.100000Z | 2019-10-17T00:00:00.100000Z | 101 | 100  |
+| 2019-10-17T00:00:00.300000Z | 2019-10-17T00:00:00.300000Z | 102 | 101  |
+| 2019-10-17T00:00:00.500000Z | 2019-10-17T00:00:00.400000Z | 103 | 102  |
+| 2019-10-17T00:00:00.600000Z | 2019-10-17T00:00:00.400000Z | 104 | 102  |
+
+The result has all rows from the `bids` table joined with rows from the `asks` table. 
+The ASOF join will look for the value in the `asks` table that has the closest
+prior to or equal to the timestamp in the `bids` table.
+
+The first row in the result set has a timestamp in the `bids` table `2019-10-17T00:00:00.000000Z`
```
jerrinot, line 227: this might be too detailed, but I find this kind of "walkthrough" helps to understand it. 
I was hesitating whether it belong to a reference manual. but then I concluded it helps to understand it.

docs/reference/sql/join.md
```diff
@@ -168,72 +171,71 @@ CROSS JOIN deserts;
...
-return the corresponding record in the other table at the closest timestamp
-**prior to** the timestamp in the first table.
-
-:::note
-
-To be able to leverage `ASOF JOIN`, both joined tables must have a
-[designated timestamp](/docs/concept/designated-timestamp/) column.
-
-:::
+More formally: `ASOF JOIN` will return all records from the left table combined with matching records from the right table.
```
amyshwang, line 178: What does "more formally" mean here? More precisely?

docs/reference/sql/join.md
```diff
@@ -168,72 +171,71 @@ CROSS JOIN deserts;
 :::
 
 ## ASOF JOIN
+`ASOF JOIN` will join two tables based on their timestamp distance.
```
amyshwang, line 174:
```suggestion
`ASOF JOIN` joins two tables based on their timestamp distance.
```

docs/reference/sql/join.md
amyshwang, line 178:
```suggestion
More formally: `ASOF JOIN` returns all records from the left table combined with matching records from the right table. 
```

docs/reference/sql/join.md
```diff
@@ -168,72 +171,71 @@ CROSS JOIN deserts;
...
+| timebid                     | timeask                     | bid | ask  |
+|-----------------------------|-----------------------------|-----|------|
+| 2019-10-17T00:00:00.000000Z | NULL                        | 101 | NULL |
+| 2019-10-17T00:00:00.100000Z | 2019-10-17T00:00:00.100000Z | 101 | 100  |
+| 2019-10-17T00:00:00.300000Z | 2019-10-17T00:00:00.300000Z | 102 | 101  |
+| 2019-10-17T00:00:00.500000Z | 2019-10-17T00:00:00.400000Z | 103 | 102  |
+| 2019-10-17T00:00:00.600000Z | 2019-10-17T00:00:00.400000Z | 104 | 102  |
+
+The result has all rows from the `bids` table joined with rows from the `asks` table. 
+The ASOF join will look for the value in the `asks` table that has the closest
```
amyshwang, line 224:
```suggestion
The ASOF join looks for the value in the `asks` table that has the closest timestamp
```

docs/reference/sql/join.md
jerrinot: yes, _more precisely_ is better than _more formally_.  thank you!

This is the idea: ASOF JOIN is not a well-known join type. It's largely specific to time-series databases. We have to educate people about it. And at the same time it's not trivial to explain it. So let's start with something simple and make it more precise gradually. This is the simplest version I could come up with (with @javier's help):
> It can join two different time-series measured at different points in time. For each value
in the first time-series, it takes the most recent value in the second time-series.

This ☝️ is supposed to give you a rough idea of what the ASOF JOIN does. It answer the question _What is it good for?_ This is how I would read it: _"ASOF JOIN is joining timestamps. and it has something to do with "recency"_. 

It does not tell you what exactly is the role of "recency". What are the rules for selecting the record from the right table. Hence the second paragraph:
> More ~formally~ precisely: `ASOF JOIN` returns all records from the left table combined with matching records from the right table. When a record from the left table has multiple matching records in the right table the ASOF JOIN selects only the record with the closest timestamp **prior or equal to** the timestamp of the left record. ASOF JOIN behaves as the LEFT OUTER JOIN with additional time filter so the right table returns at most one record per each record from the left side.

This ☝️ is more precise than the shorter version above. At the same time, it's a mouthful. I wouldn't want to read this as a very first intro to ASOF JOINs. So the short intro tells you: "Why would you want to even consider it". It's basically a motivation. The 2nd paragraph tells you: "How exactly it works?" - giving you the exact rules - when you actually feel ASOIF is what you need.

Does it make any sense to you?:)

docs/reference/sql/join.md
```diff
@@ -169,74 +172,76 @@ CROSS JOIN deserts;
...
+result set has a timestamp in the `bids` table `2019-10-17T00:00:00.600000Z` and
+the closest earlier timestamp in the `asks` table is also `2019-10-17T00:00:00.400000Z`
+and the join returns this row.
 
-```questdb-sql
-SELECT bids.ts timebid, bid, ask
-FROM (bids timestamp(ts))
-ASOF JOIN (asks timestamp (ts));
-```
+## Matching Clause
```
amyshwang, line 240:
```suggestion
### Matching Clause
```
I think it should be level 3?

amyshwang, line 240:
```suggestion
### Matching Clause
```
I think it should be level 3?

docs/reference/sql/join.md
```diff
@@ -169,61 +171,139 @@ CROSS JOIN deserts;
...
+| 2019-10-17T00:00:00.400000Z | 502 | AAPL       |
+| 2019-10-17T00:00:00.700000Z | 200 | IBM        |
+
+Notice how both tables have a new column `stock` that stores the stock name.
+The `ON` clause allows you to match the value of the `stock` column in the `bids` table with
+that in the `asks` table:
 
 ```questdb-sql
-SELECT bids.ts timebid, bid, ask
+SELECT bids.stick stock, bids.ts timebid, asks.ts timeask, bid, ask
```
amyshwang, line 269:
```suggestion
SELECT bids.stock stock, bids.ts timebid, asks.ts timeask, bid, ask
```

src/css/_global.css
```diff
@@ -442,3 +442,15 @@ html[data-theme='light'] .react-toggle-track {
 .transition-node-enter-active { opacity: 1; transition: opacity 200ms; }
 .transition-node-exit { opacity: 1; }
 .transition-node-exit-active { opacity: 0; transition: opacity 200ms; }
+
+.pink-table table {
+  --ifm-table-background: var(--palette-pink);
```
amyshwang, line 447: where it says 'pink' is it defined to our QuestDB logo colour?

amyshwang: The output uses these two colours
Screenshot 2023-02-21 at 17 22 56

But ideally we want #A2356D and #366994
Also, is it possible to use different colour for different columns?
Screenshot 2023-02-21 at 17 25 05
In short, we are wondering if it is possible to avoid using image files for tables...

docs/reference/sql/join.md
```diff
@@ -169,61 +171,143 @@ CROSS JOIN deserts;
...
-a table is created as ordered by time, the order of records is enforced and the
-timestamp column name is in the table metadata. `ASOF` join will use this
-timestamp column from metadata.
+```questdb-sql
+SELECT bids.ts timebid, asks.ts timeask, bid, ask
+FROM bids
+ASOF JOIN asks;
+```
+This is the JOIN result:
+![The query result with pink background for the bids table and blue for the asks table](/img/docs/reference/sql/asof-result.png)
```
amyshwang, line 192: This image and line182 are created as png files because I thought it was not possible to add different background colours to a markdown table.

docs/reference/sql/join.md
```diff
@@ -169,61 +171,143 @@ CROSS JOIN deserts;
...
+The above query returns these results:
+
+| timebid                     | timeask                     | bid | ask  |
+| --------------------------- | --------------------------- | --- | ---- |
+| 2019-10-17T00:00:00.000000Z | NULL                        | 101 | NULL |
+| 2019-10-17T00:00:00.100000Z | 2019-10-17T00:00:00.100000Z | 101 | 100  |
+| 2019-10-17T00:00:00.300000Z | 2019-10-17T00:00:00.300000Z | 102 | 101  |
+| 2019-10-17T00:00:00.500000Z | 2019-10-17T00:00:00.400000Z | 103 | 102  |
+| 2019-10-17T00:00:00.600000Z | 2019-10-17T00:00:00.400000Z | 104 | 102  |
+-->
```
amyshwang, line 232: This is the markdown source of the table. I keep the source in case we need to update the content in the future - images are not the best source.

docs/reference/sql/join.md
```diff
@@ -169,61 +171,143 @@ CROSS JOIN deserts;
...
+### Using `ON` for matching column value
+
+An additional `ON` clause can be used to join the tables based on the value of a selected column.
+
+The query above does not use the optional `ON` clause. If both tables store
+data for multiple stocks, `ON` clause provides a way to find asks for bids with
+matching stock value.
+
+Table `bids` (the left table):
+<div class="pink-table">
```
amyshwang, line 243: Here @jerrinot is exploring whether it is possible to keep the source to markdown. 
However ideally we want to apply the colours to the tables from L207. Basically using markdown to reproduce what I had in png files. https://github.com/questdb/questdb.io/pull/1406/files#r1113370539

amyshwang (commented):
Adding comments about table colours.

src/css/_global.css
```diff
@@ -442,3 +442,25 @@ html[data-theme='light'] .react-toggle-track {
...
+  --ifm-table-stripe-background: var(--palette-pink);
+  --ifm-table-head-background: var(--palette-pink);
+}
+
+.blue-table table {
+  --ifm-table-background: var(--palette-pale-blue);
+  --ifm-table-stripe-background: var(--palette-pale-blue);
+  --ifm-table-head-background: var(--palette-pale-blue);
```
amyshwang, line 455: Could we also define blue table to `background: #943d6b;`

src/css/_global.css
```diff
@@ -442,3 +442,25 @@ html[data-theme='light'] .react-toggle-track {
 .transition-node-enter-active { opacity: 1; transition: opacity 200ms; }
 .transition-node-exit { opacity: 1; }
 .transition-node-exit-active { opacity: 0; transition: opacity 200ms; }
+
+.pink-table table {
+  --ifm-table-background: var(--palette-pink);
+  --ifm-table-stripe-background: var(--palette-pink);
+  --ifm-table-head-background: var(--palette-pink);
```
amyshwang, line 449: Could we define pink table to #943d6b?

PR #1413, opened 2023-02-22
chore(docs) document year as a valid sampling unit

jerrinot:
Year is a supported sampling unit too:
- https://github.com/questdb/questdb/blob/526a879ab825adfe1f8f77e8757164f298deb301/core/src/main/java/io/questdb/griffin/engine/groupby/TimestampSamplerFactory.java#L124
- https://github.com/jerrinot/questdb/blob/503beb7b6c4ae8312ec5d1b9a969034c0f35ca41/core/src/main/java/io/questdb/griffin/SqlParser.java#L128
- https://github.com/questdb/questdb/blob/526a879ab825adfe1f8f77e8757164f298deb301/core/src/test/java/io/questdb/griffin/engine/groupby/SampleByTest.java#L5921


puzpuzpuz (approved):
Nice catch!

PR #1418, opened 2023-02-23
chore(docs): change license of the website and docs to CC BY-SA 4.0

ideoma:
Change the license Apache 2.0 to CC BY-SA 4.0 used by Wikipedia which is more appropriate for the website content.


amyshwang:
Postponed as discussed with @tris0laris

PR #1425, opened 2023-02-27
WAL documentation proofreading

sklarsa:


docs/concept/write-ahead-log.md
```diff
@@ -58,19 +59,19 @@ table:
...
 :::note
 
-We are working to reduce the limitations.
+We are working hard to reduce the below limitations.
 
 :::
 
 For a WAL table, the following existing operations may have different behaviors
-to a non-WAL table:
+than a non-WAL table:
```
amyshwang, line 67: I think either 'different to' or 'different from' work better than different 'than'?

amyshwang (commented):
A small question added.

docs/concept/write-ahead-log.md
sklarsa: I was talking to my wife about that one lol. I think different from is correct

amyshwang (approved):

PR #1444, opened 2023-03-07
feat: make blog and docs section follow OS theme (light | dark)

ideoma:


argshook (changes requested):
I'm not sure theme should be switching automatically without user intent. With this PR, when navigating between pages theme changes on its own and this can resemble a bug.

For example `docs` can be shown with `light` theme, but visiting `pricing` or `get questdb` theme suddenly can switch to `dark`. When navigating back the theme can switch again. This behavior doesn't feel right to me.

User can click on theme toggle switch at the top and this would set the theme. After clicking it, theme doesn't switch anymore when navigating between pages. But this then negates the need of code in this PR.

Moreover, for mobile users theme toggle button is shown under a menu, it's not immediately visible. They might not find (or not that they need to find) theme toggle switch in order to stop automatic theme switching.

If user has `light` or `dark` theme preference, they can use that toggle switch and the setting is saved and respected between visits.

Or, if the intent of this PR is to make website respect the OS light/dark setting, then we can change docusaurus setting [`respectPrefersColorScheme: false`](https://github.com/questdb/questdb.io/blob/c854282f7dc50488403cc05386e368860f8da27d/docusaurus.config.js#L94) to `true`. In this case there should be no need of any additional code.

Another point why i wouldn't want this PR merged is that we would need to maintain not only `theme` in `localStorage` but also `theme-main`.

As of now i fail to understand how this change would benefit the user.

ideoma:
This is the only "dark" mode website in the database world I've seen recently. Click on "Docs" or "Blog" to see what it does 
https://materialize.com/

argshook:
> This is the only "dark" mode website in the database world I've seen recently. Click on "Docs" or "Blog" to see what it does 
> 
> https://materialize.com/
> 
>  

Thanks for sharing an example!

I think it's okay to let user change the theme (as is the case now) and we can improve first-time experience by changing 
`respectPrefersColorScheme` to `true` in `docusaurus.config.js`.

But to switch theme automatically without user intent just doesn't feel right to me.

Although i don't agree, i'm not here to block, so i will approve the PR.

argshook (approved):

amyshwang:
I checked the example and I personally found it very strange that the theme just changes - my eyes don't like it at all. I am also not sure about making the theme the same as the OS setting - I do like the theme colour I chose for each application independent of my OS theme.
But it is just my own preference.

amyshwang:
Closing this PR as we have decided to leave things as they are.

PR #1453, opened 2023-03-10
chore(docs): add Maven coordinates for the Flink connector

jerrinot:


amyshwang (commented):
I'd update the link in L51 because it is hard coded. We could use this: https://repo1.maven.org/maven2/org/questdb/flink-questdb-connector/

Alternatively in L51 I'd change the anchor to 'Download' as the link download the jar file.

jerrinot:
I considered that. but then I think it's better to keep it as it is. for the following reasons:
1. the tutorial was tested with connector version 0.2. it would be a shame if it stopped working because of some incompatibility in a next version
2. right now you can right-click the link from the tutorial and paste it to `wget`.  with https://repo1.maven.org/maven2/org/questdb/flink-questdb-connector/ you would have to write something like "download the most recent connector version from...". I am not sure if that's great for a tutorial. a tutorial should do as much hand-holding as possible? I think.

amyshwang (approved):
Thank you!

PR #1459, opened 2023-03-14
improve tutorials tags page

argshook:
This PR introduces ability to `pin` blog posts in order to show them at the top of a tags page.

A tags page is one like https://questdb.io/blog/tags/tutorial

Blog posts that include `pinned` tag in their front-matter (the config section at the top of `*.mdx` files) will be:

* shown at the top of the tags page
* excluded from the "all posts" list

In addition, this PR updates slightly the UI of "Popular topics" and "see all topics" areas. The intent is to reduce verbosity and visual "weight" of them.

# Before `/blog/tags/tutorial`

Screenshot 2023-03-15 at 12 43 11

# After `/blog/tags/tutorial`

> please note there currently are only 4 pinned tutorial posts. This can be easily changed by adding/removing `pinned` tag on selected tutorial posts. Please let me know if current ones should be changed.

Screenshot 2023-03-15 at 12 43 21

# Before `/blog`

Screenshot 2023-03-15 at 12 49 07

# After `/blog`

Screenshot 2023-03-15 at 12 49 24


src/theme/BlogListPage/index.tsx
```diff
@@ -62,6 +64,22 @@ const prioritizedTags: ChipProps["items"] = [
   permalink: `/blog/tags/${tag.replace(/ /g, "-")}`,
 }))
 
+const pinnedPostsTitle = (tag: string) => {
+  const map: Record<string, string> = {
+    tutorial: "Featured QuestDB tutorials",
```
amyshwang, line 69: I noted that the font of this title is bigger than other titles. 
1. The font size is in keeping with the blog landing page where the first blog post title has the same enlarged title
2. However on the tutorial landing page, this title feels much bigger. Essentially it is a section title, not an article title so should we align it with the font size of other sections? (Popular topics etc):
image

tris0laris (approved):

PR #1467, opened 2023-03-16
fix(docs:) aggregation by hourly time buckets example

flozdra:
The SAMPLE BY keyword was missing in this example query


amyshwang (approved):
Thank you @flozdra for your contribution!

flozdra:
Hi! I just noticed that I rounded the timestamp to the minute and not to the hour 🤦

PR #1535, opened 2023-04-11
feat(blog): explain with basic scan types

bziobrowski:


blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,436 @@
...
+
+ | QUERY PLAN                                            |
+|-------------------------------------------------------|
+| DataFrame                                             |
+| &nbsp;&nbsp;&nbsp;&nbsp;Row forward scan              |
+| &nbsp;&nbsp;&nbsp;&nbsp;Frame forward scan on: trades |
+
+As expected, optimizer recognized that `order by timestamp` can be implemented simply by forward scanning the table, no sorting required.
+In `EXPLAIN` output a Table forward scan is represented on table level - `Frame forward scan` and partition level - `Row forward scan`.
+
```
bziobrowski, line 84: @amyshwang I included "backward" version for table scan only because I felt it would be too repetitive . 
Let me know if you think we should include it for all scan types .

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,436 @@
...
+```questdb-sql
+CREATE TABLE trades (
+ symbol SYMBOL,
+ side SYMBOL,
+ price DOUBLE,
+ amount DOUBLE,
+ timestamp TIMESTAMP
+) timestamp (timestamp) PARTITION BY DAY WAL;
+```   
+
```
bziobrowski, line 51: I'm mixing terms here a bit because the common term is "Table scan" or "Full table scan" while explain's output shows "Frame forward scan" . Do you think it's ok as is or should I change it

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,436 @@
...
+>
+Photo by <a href="https://unsplash.com/@benjaminlehman">Benjamin Lehman</a> via{" "}
+<a href="https://unsplash.com">Unsplash</a>
+</Banner>
+
+Welcome to another post on sql performance tuning !
+This time we'll explore the various scan types (a.k.a. access methods) QuestDB supports.
+It is necessary to understand them before tackling more complex queries. 
+
+What are scan types  ?
```
amyshwang, line 32: "scan type" or "access mode" does not seem to be a mainstream keywords. 
I googled some terms like database scan but it does not convey the same thing. Is there a better term widely used that we could use here?

If not, we could use a cluster of strong keywords to describe the concept in the header.

We need a good level twl header here before jumping into each type.

blog/2023-04-20-explain-basic-scans.md
amyshwang: I think it is important to be logical so:

- (full) table scan
    - Frame forward scan
    - Frame backward scan

This is very clear.
What we should consider more is to add more descriptive headers because Frame forward/backward scan is QuestDB terms.

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,436 @@
...
+| &nbsp;&nbsp;&nbsp;&nbsp;DataFrame                                                   |
+| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Row forward scan                    |
+| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Frame forward scan on: trades       |
+
+Rewriting predicate as `timestamp between '2023-01-19T23:59:00.000000Z' and '2023-01-20T23:59:00.000000Z'` makes query use interval filter.  
+Lesson - keep designated timestamp predicates clean and simple!
+
+#### Interval backward scan
+
+Same as forward scan but runs from last row of last interval to first row of first interval.
```
amyshwang, line 180: This applies elsewhere too - I'd use complete sentences.

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,436 @@
...
+produces :
+
+| QUERY PLAN     |
+|----------------|
+| Empty table    |
+
+If predicate is too complex (especially if designated timestamp is used as function argument), engine will fall back to default table scan with filter, e.g. 
+
+```questdb-sql
+explain select * from trades
```
amyshwang, line 162: Let's use the correct syntax and capitalisation?

amyshwang (changes requested):
I think we need to:
- Contextualise how we are using EXPLAIN to showcase scan types
- Headers may need to be more descriptive in order to tell a story (and using our focused keywords
- Article title is not using solid keywords
- SQL query capitalisation

blog/2023-04-20-explain-basic-scans.md
bziobrowski: I meant that diagram-wise . Do you think I should add separate diagram for interval backward scan, index backward scan, ... ?

blog/2023-04-20-explain-basic-scans.md
bziobrowski: Maybe a bit of explanation would do ? 
The problem is that our terms don't entirely match the standard ones and EXPAIN's output is custom . 
Normally "Full table scan" is done from start to end of table because data is not ordered and direction doesn't matter . Qdb tables that have designated timestamp are ordered by it so direction matters a lot . 

Interval scan : 
- conceptually is a sub-type of full table scan 
- can be done in forward or backward direction
- is specific to qdb . 

Index scans are more straightforward, expect that our indexes aren't the usual b-trees but hash-map-like .

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,436 @@
...
+</Banner>
+
+Welcome to another post on sql performance tuning !
+This time we'll explore the various scan types (a.k.a. access methods) QuestDB supports.
+It is necessary to understand them before tackling more complex queries. 
+
+What are scan types  ? 
+Scan type is an algorithm used to find and access data. 
+Imagine binary search on sorted data, but with partitions, indexes and column data files instead of arrays.   
+There are two basic scan types in QuestDB, table and index scan.
```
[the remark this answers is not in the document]
bziobrowski: If we use standard terms then interval scan is like a sub-type of full table scan but if we use ours then we can organize it like this : 

- Table scans 
  - Frame forward/backward scan 
  - Interval forward/backward scan 
- Index scans 
  - Index forward/backward scan in index order 
  - Index forward/backward scan in table order 

What do you think ?

blog/2023-04-20-explain-basic-scans.md
amyshwang: I guess we could do a diagram for each scan type with both directions added to the diagrams? For example:

Screenshot 2023-04-13 at 12 46 03

blog/2023-04-20-explain-basic-scans.md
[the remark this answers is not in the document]
amyshwang: We may not want to keep the headers exactly the same but I think overall this structure works better:
- Table scans
   - Frame (forward/backward) scan
   [diagram indicating both directions]
      - text and examples
   - Interval (forward/backward) scan
      [diagram indicating both directions]
      - text and examples
- Index scans
   - Index (forward/backward) scan in index order
        [diagram indicating both directions]
      - text and examples
  - Index (forward/backward) scan in table order
     [diagram indicating both directions]
      - text and examples

amyshwang:
I pushed commit `[e57622b](https://github.com/questdb/questdb.io/pull/1535/commits/e57622b378fe678819e4c9fc97d3d64bb3cbd6de)` for editorial changes - it is important that the changes do no alter the meaning.

I think this will then allow us to focus on terms and headers to use.

blog/2023-04-20-explain-basic-scans.md
amyshwang: UUnderstood.
But is this clear in the article?

> expect that our indexes aren't the usual b-trees but hash-map-like

blog/2023-04-20-explain-basic-scans.md
amyshwang: I moved the existing diagrams to places that I think suitable and they can be combined to one diagram for both directions.

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,447 @@
...
+| FilterOnValues symbolOrder: asc                                                            |
+| &nbsp;&nbsp;&nbsp;&nbsp;Cursor-order scan                                                  |
+| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Index backward scan on: id deferred: true  |
+| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;filter: id='X'                 |
+| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Index backward scan on: id deferred: true  |
+| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;filter: id='Y'                 |
+| &nbsp;&nbsp;&nbsp;&nbsp;Interval forward scan on: pos                                      |
+| &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;intervals: [static=[1675209600000000,1675295999999999] |
+
+Note that while technically we're doing an interval forward scan, it doesn't really matter because there's just one interval to scan. While this makes sense with a single partition, it doesn't extend to the case with multiple partitions. That's because, on a higher level, the SQL engine iterates over partitions, and in doing so it mixes the order.
```
amyshwang, line 341: I don't understand this because the plan says "Index backward scan"...Why is it technically a forward scan? Because it's sequential to the key?

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,447 @@
...
+#### Index order
+
+Index scan with multiple values in index order scans table rows using row ids associated with the first index key value, followed by the second and the third. The scan continues until reaching the last index value. It might be used to avoid sorting at the price of potentially more random memory and disk accesses.
+
+<Banner
+  alt="Index forward scan in index order."
+  height={433}
+  src="/img/blog/2023-04-20/index_forward_scan_index_order.svg"
+  width={650}
+/>
```
amyshwang, line 321: I forgot to move this to L313

blog/2023-04-20-explain-basic-scans.md
bziobrowski: For table scans it should be fine but wouldn't we need to change indices on index scan diagrams to show rows being read in backward direction ?

blog/2023-04-20-explain-basic-scans.md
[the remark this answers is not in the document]
amyshwang: Apologies - the structure I made in [e57622b](https://github.com/questdb/questdb.io/pull/1535/commits/e57622b378fe678819e4c9fc97d3d64bb3cbd6de) is what I think would work - I did keep the single key part!

blog/2023-04-20-explain-basic-scans.md
amyshwang: For table order and index order scan, I don't think the arrows for forward scan makes sense because (if I understand it correctly) you have numbered the scan sequence.

We could use the diagram, remove the arrow and add:

Index forward scan -multiple keys, table order]: scanning from (1) to (6) sequentially
Index forward scan -multiple keys, table order]: scanning from (6) to (1) in decreasing order

blog/2023-04-20-explain-basic-scans.md
amyshwang: You did say:
> In the EXPLAIN output, a table forward scan is represented on the table level as Frame forward scan, and on the partition level, Row forward scan.

But it's under Table scan and by the time I am at Index scan, I don't remember that piece of info anymore @@ 

Maybe moving that line to the intro, before "Table scan" and after:

> QuestDB supports two main scan types, table and index scan. The difference is that the former touches table data directly, while the latter goes through the index first. Both types include sub-types and variants.

So in the intro we talk about:
- scan types
- scan levels

blog/2023-04-20-explain-basic-scans.md
bziobrowski: The arrow is not necessary, it just shows the scan direction on partition level .

blog/2023-04-20-explain-basic-scans.md
amyshwang: I think this paragraph has not been updated and is still a bit confusing.

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,440 @@
...
+CREATE TABLE pos (
+ time TIMESTAMP,
+ id SYMBOL INDEX,
+ lat DOUBLE,
+ lon DOUBLE,
+ geo6 GEOHASH(6c),
+ geo12 GEOHASH(12c)
+) timestamp (time) PARTITION BY DAY;
+```
+
```
amyshwang, line 214: Is it right to insert a diagram here? It should be in "Index scan with a single key"?

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,440 @@
...
+Both have pros and cons, so let's look at a simple example.
+
+#### Table order
+
+<Banner
+  alt="Index forward scan in table order."
+  height={433}
+  src="/img/blog/2023-04-20/index_scan_table_order.svg"
+  width={650}
+/>
```
amyshwang, line 286: The arrow only shows a forward scan. I wonder if my previous suggestion [here](https://github.com/questdb/questdb.io/pull/1535/files/57ea3e9071d825d181ee4297adca22c7993694c6#r1166356099) still applies?
> 
> We could use the diagram, remove the arrow and add:
> 
> Index forward scan -multiple keys, table order]: scanning from (1) to (6) sequentially
> Index forward scan -multiple keys, table order]: scanning from (6) to (1) in decreasing order

And the same for the index order diagram, where the arrow does not make sense.

amyshwang (commented):
Further comments about diagrams.

blog/2023-04-20-explain-basic-scans.md
bziobrowski: I removed arrows but netlify still shows old diagrams.

amyshwang: Probably due to the cache.
I am not sure if it is obvious that the number in parenthesis indicates the order though. It took me a while to figure that out and I am still not sure if that's correct. Could you add some text to explain it?

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,575 @@
...
+
+This output would take too much space.
+
+Therefore, QuestDB's EXPLAIN command shows:
+
+- The direction of Data Frame (a piece or the whole partition) iteration, e.g.
+  `Frame forward scan`
+- The direction of Row iteration within a Data Frame, e.g. `Row forward scan`
+- A list of scan boundaries in `Interval scan`, e.g.
+  `intervals: [("1970-01-01T00:00:00.000000Z","1970-01-01T23:59:59.999999Z")]`
```
amyshwang, line 142: I feel this important info should be in the intro as it is not just for Table scan?

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,575 @@
...
+We have introduced the basic scan types available in QuestDB:
+
+- Frame scan
+- Interval scan
+- Index scan (in table and index order)
+
+We have also demonstrated the available optimizations for tables with the
+designated timestamp.
+
+We hope you enjoy this tutorial.
```
amyshwang, line 573: Could we add an ending highlighting the benefit of using EXPLAIN to refine SQL queries?

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,575 @@
+---
+title: "EXPLAIN scan types"
```
amyshwang, line 2: We will need to change the title and slug

docs/reference/sql/explain.md
```diff
@@ -261,4 +261,5 @@ methods.
 
 This section includes links to additional information such as tutorials:
 
-- [SQL Performance Tuning: Introducing EXPLAIN](/blog/explain-sql-query-plan/)
+- [EXPLAIN Your SQL Query Plan](/blog/explain-sql-query-plan/)
+- [EXPLAIN Basic Scans ](/blog/explain-sql-index-table-scan/)
```
amyshwang, line 265: This needs to change depending on the final title

blog/2023-04-20-explain-basic-scans.md
bziobrowski: I put it here because it makes them most sense with an example .

blog/2023-04-20-explain-basic-scans.md
bziobrowski: Added in latest commit .

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -0,0 +1,583 @@
...
+  src="/img/blog/2023-04-20/index_scan_index_order.svg"
+  width={650}
+>
+  Index forward scan in index order: scanning from (1) to (6) 
+  Index backward scan in index order: scanning from (6) to (1)
+</Banner>
+
+Index scans with multiple keys in index order scan table rows using row ids
+associated with the first index key, followed by those associated with the
+second index key, and so on.
```
amyshwang, line 413: The description and the diagram seem to be contradictory to each other.

Screenshot 2023-04-21 at 14 31 49


>Index scans with multiple keys in index order scan table rows using row ids
associated with the first index key, followed by those associated with the
second index key, and so on.

The diagram shows scanning table rows using row ids
associated with the first index key in the first partition, followed by the table row ids with the second index key in the same partition. The scan then continues to the second partition.

amyshwang (commented):
One final question!

blog/2023-04-20-explain-basic-scans.md
bziobrowski: I guess the explanation on two levels of iteration/scanning isn't clear enough . 
Indexes are partitioned  and scanning on the top level happens on partitions or intervals and for each such 'Frame' runs a  row or index scan .

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -348,9 +347,8 @@ scan with an Interval scan.
 
 Things get more interesting when there's more than one index key to scan. As
 shown in the index scan diagram above, indexes are partitioned. When QuestDB
-iterates over a partition or interval list, it conducts an index scan for each
-list followed by a scan for each single key value, with the output still in
-timestamp order.
+iterates over a partition or interval list and conducts an index scan for each
+partition or interval, the output still in timestamp order.
```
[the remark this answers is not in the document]
amyshwang:
```suggestion
partition or interval, with the output still in timestamp order.
```

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -81,9 +82,9 @@ A query selecting all columns amounts to a traditional Full Table Scan. A query
 selecting a single/a few columns from a wide table might read just a few percent
 of the table's data.
 
-Frame forward scans start at the first row of the oldest partition and stop at
-the last row of the latest one. Differentiating between forward and backward
-scans only makes sense for tables with a designated timestamp because they store
+Frame forward scans start at the first row of the oldest partition (one with lowest timestamp partition key value)
```
amyshwang, line 85: the one with the lowest timestamp, and the one with the highest timestamp in L86

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -139,7 +140,7 @@ Therefore, QuestDB's EXPLAIN command shows:
 - The direction of Data Frame (a piece or the whole partition) iteration, e.g.
   `Frame forward scan`
 - The direction of Row iteration within a Data Frame, e.g. `Row forward scan`
-- A list of scan boundaries in `Interval scan`, e.g.
+- A list of scan boundaries in `Interval scan`, e.g.\
```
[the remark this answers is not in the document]
amyshwang:
```suggestion
- A list of scan boundaries in `Interval scan`, e.g.
```

blog/2023-04-20-explain-basic-scans.md
```diff
@@ -237,11 +238,9 @@ WHERE dateadd('m', 1, timestamp) in '2023-01-20';
 | &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Row forward scan                    |
 | &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Frame forward scan on: trades       |
 
-Rewriting the `WHERE` condition to
-`timestamp BETWEEN '2023-01-19T23:59:00.000000Z' AND '2023-01-20T23:59:00.000000Z'`
-makes the query use the interval scan.
-
-Lesson - keep the conditions on the designated timestamp clean and simple!
+Rewriting the `WHERE` condition to\
+`timestamp BETWEEN '2023-01-19T23:59:00.000000Z' AND '2023-01-20T23:59:00.000000Z'`\
```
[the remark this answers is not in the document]
amyshwang:
```suggestion
Rewriting the `WHERE` condition to
`timestamp BETWEEN '2023-01-19T23:59:00.000000Z' AND '2023-01-20T23:59:00.000000Z'`
```

amyshwang (changes requested):

blog/2023-04-20-explain-basic-scans.md
amyshwang: Might it be better to update this paragraph too?
```suggestion
Multiple-key-index-order scans iterate one partition after another,  scanning table rows using row ids associated with the first index key, followed by those associated with the second index key, and so on.
```

amyshwang (approved):
Thank you @bziobrowski for your patience!

PR #1536, opened 2023-04-11
fix(doc): fix within operator doc.

bziobrowski:


amyshwang (changes requested):
Some minor editorial comments

PR #1542, opened 2023-04-12
chore(*): add tags on docs sidebar

argshook:
- add tags to sidebar items
- adjust sidebar tag colors


amyshwang:
I can think of tagging https://questdb.io/docs/concept/write-ahead-log/ as "New" but if we are using tags for "New" and "Popular", I would not use it to indicate part of the original title - "COPY SQL" and "REST API" as it might be a bit confusing?

Having said that "COPY SQL" and "REST API" rendered as tags look nice. I can also see us using tags in this way. But I don't think we should mix the usages...Maybe I am overthinking...

argshook:
> I can think of tagging https://questdb.io/docs/concept/write-ahead-log/ as "New" but if we are using tags for "New" and "Popular", I would not use it to indicate part of the original title - "COPY SQL" and "REST API" as it might be a bit confusing?
> 
> Having said that "COPY SQL" and "REST API" rendered as tags look nice. I can also see us using tags in this way. But I don't think we should mix the usages...Maybe I am overthinking...

I'd say let's push and observe. This is a small tweak that doesn't have impact on SEO or similar areas and is very easy to change.

amyshwang (approved):

PR #1551, opened 2023-04-13
Overflowing content fixed at use cases

saikrishnatbijil:
fixed overflowing content at use cases page under
**before :**
Screen Shot 2023-04-13 at 2 38 59 PM
**after:**
Screen Shot 2023-04-13 at 2 42 58 PM

PR #1556, opened 2023-04-13
use cases page overflow on monitoring section

saikrishnatbijil:
this is follow up of #1551
requested changes added made it responsive @argshook. 
now:
Screen Shot 2023-04-13 at 11 31 39 PM


saikrishnatbijil:
is there a problem with this way of making it responsive @argshook

src/pages/use-cases.tsx
```diff
@@ -190,7 +190,7 @@ const Monitoring = () => (
       <img
         src="/img/pages/use-cases/real-time-analytics-jumbo.svg"
         alt="An illustration of real-time analytics and monitoring"
-        width="630"
+        width="55%"
```
argshook, line 193: unfortunately this is not good enough :(

in HTML it's preferred to keep width as exact pixels, because it helps prevent layout shift while the image is loading.

In addition, changing only one image to percent width is not enough as there are more images on the page.

However, you can achieve a good layout with these changes in `src/css/use-cases/use-case.module.css`:

1. line `30`:
```
.use-case__image {
  max-width: 50%;
}
```

2. somewhere inside media query, around line `34`:

```
.use-case__image {
  max-width: 50%;
}
```

argshook (changes requested):

src/pages/use-cases.tsx
saikrishnatbijil: ok i will complete this no probs. Thanks for the support

saikrishnatbijil:
while fixing the images issue i found another overflow issue on the use_case_customer i fixed that too.. @argshook

src/css/use-cases/use-case.module.css
```diff
@@ -62,6 +66,9 @@
     display: flex;
     align-items: center;
   }
+  .use-case__image {
+    width: 70%;
+  }
```
argshook, line 71:
```suggestion
```

i wouldn't add this, as we don't need to focus illustration so much on narrow screen. IMO single 50% is good enough

src/components/UseCaseQuotes/use-case-customers.module.css
```diff
@@ -1,8 +1,9 @@
 .use-case-customers {
+  max-width: 100%;
   padding: 2.6rem 0 0 0;
   border-radius: 16px;
   background: var(--theme-card-bg-color);
-  flex: 1;
+  flex: 0.9;
```
argshook, line 6:
```suggestion
  flex: 1;
```

unsure why `0.9` is used, i would keep `1`. It looks just as good

argshook (changes requested):

argshook (approved):

argshook:
Thanks for contributing!

PR #1559, opened 2023-04-14
chore(docs): changing warning to note

amyshwang:


amyshwang:
Need to look at all the blocks in a holistic way.

PR #1566, opened 2023-04-15
fixed overflowing tables at quest db cloud page

saikrishnatbijil:
fixed overflowing tables at quest db cloud page. 
**before:**
Screen Shot 2023-04-15 at 4 45 23 PM
**after:**
Screen Shot 2023-04-15 at 4 50 28 PM
At smaller screen sizes (bellow 400px) the tables are overflowing due to some padding issue. i fixed it by removing the extra padding

### DIFF
--- a/src/modules/cloud/FeatureTable/styles.module.css
+++ b/src/modules/cloud/FeatureTable/styles.module.css
@@ -19,12 +19,6 @@
   font-weight: var(--ifm-font-weight-bold);
 }
 
-@media (max-width: 400px) {
-  .root thead th {
-    padding: 0rem;
-  }
-}
-
 .root thead th:first-child {
   text-align: left;
   width: var(--first-col-width);
--- a/src/modules/cloud/FeatureTable/styles.module.css
+++ b/src/modules/cloud/FeatureTable/styles.module.css
@@ -19,6 +19,12 @@
   font-weight: var(--ifm-font-weight-bold);
 }
 
+@media (max-width: 400px) {
+  .root thead th {
+    padding: 0rem;
+  }
+}
+
 .root thead th:first-child {
   text-align: left;
   width: var(--first-col-width);
--- a/src/modules/cloud/CompareFeatures/index.tsx
+++ b/src/modules/cloud/CompareFeatures/index.tsx
@@ -170,7 +170,7 @@ export const CompareFeatures = () => {
         Features
       </Section.Title>
 
-      <Section>
+      <Section noGap>
         <div className={style.tables}>
           <FeatureTable title="Core features" items={CoreFeaturesItems} />
 


src/modules/cloud/FeatureTable/styles.module.css
```diff
@@ -19,6 +19,12 @@
   font-weight: var(--ifm-font-weight-bold);
 }
 
+@media (max-width: 400px) {
+  .root thead th {
+    padding: 0rem;
+  }
+}
+
```
argshook, line 27: This is a non ideal solution and there are a couple reasons why, i'll try to explain below.

we aim for designs with mobile-first approach in mind. This means
implementing design should begin with a good looking narrow screen, then
expand and adjust, then again expand and adjust and so on, until website
looks good on any width.

however, this code works in opposite direction: it tries to "squeeze"
the UI that looks good on wide screen into a narrow screen.

If mobile-first approach is implemented properly, then CSS should have
only `min-width` queries (not `max-width`!). This way by default CSS
is written for narrow screen (no media query), and when screen is expanded, `min-width` query
applies and adds new rules on top.

This is key difference to realize in order to properly work with mobile-first approach.

So, how should mobile-first be implemented in this specific case?

1. line `17` has `padding: 1rem`. It applies to all screen sizes,
   including narrow screens. As you now know, this is not good for
   narrow screen! Thus, this padding should become `padding: 0` on line
   `17`.
2. then, line `22` should use `min-width` instead of `max-width`,
   because we now want to add new rules for wider screen. Therefore:
   ```
   @media (min-width: 400px) {
      .root thead th {
        padding: 1rem;
      }
   }
   ```

This way we have a proper mobile-first approach. We can continue
expanding screen and see if it still looks good. If at some width it
doesn't, we would add another media query with another breakpoint of
`min-width`.


# However!

This particular issue can be solved by adding `noGap` property on
`<Section>` in
[src/modules/cloud/CompareFeatures/index.tsx](https://github.com/saikrishnatbijil/questdb.io/blob/2886a3c03e7017211d602d5b932eb3fe036f82a5/src/modules/cloud/CompareFeatures/index.tsx#L173):

`<Section noGap>`

And, that's it!

So in this case there's no need to do any CSS changes at all, but i
still wanted to share the mobile-first approach idea, if you'd choose to
contribute more.

My big comment here is also to expand on [this
comment](https://github.com/questdb/questdb.io/pull/1568#issuecomment-1511059247).
We're a small team and can't get back to all contributions quick. But,
if contributions are of good quality, then it's a pleasure to merge
  them.

How can you know if contribution is of good quality? This is where the
"more systematic way" can help. For instance, if the problem you'd like
to solve is "website looks bad on narrow mobile screens", it's better to
address in bulk, rather than fixing items one by one.

That's because when addressing in bulk, you might see a repeating
pattern which could be solved in a more generic way (instead of, for
example, small media-queries all around).

> this is not a golden rule! It's possible that doing a bulk change might
become a "mega PR" which is hard to review and merge. The challenge is
to find a nice balance. 

Bear in mind that when browsing code you might see that there're different
patterns and ideas. I'm talking here about mobile-first but it's not
implemented properly everywhere in the code.

This is what we want to clean up, with every contribution.

---

I hope this doesn't discourage you from contributing, I really appreciate your time and effort!

argshook (changes requested):

src/modules/cloud/FeatureTable/styles.module.css
saikrishnatbijil: Thank you i will be more conscious on my other contributions from now

argshook: Please don't let my long texts discourage you, i think you're doing great!

saikrishnatbijil:
@argshook i have done the changes it looks much better. Thank you for the suggestion

argshook (approved):

PR #1571, opened 2023-04-17
chore(*): cleanup dependencies

argshook:
- disable `react/no-unescaped-entities`
- remove unused `clean-css-cli` dep
- removeu unsed `hast-util-is-element` dep
- remove unused dependencies and move dev deps to `devDependencies`


.npmrc
```diff
@@ -0,0 +1 @@
+cache=node_modules/.cache
```
argshook, line 1: this bad boy does:

```
12:13:36 PM: Successfully restored: node_modules/.cache ... 1134 files in total.
```

previously, every build was just:

```
12:26:32 PM: A cache of 'node_modules/.cache' doesn't exist (yet).
```

insmac (approved):
@argshook Is 4.0.5 good enough for TS, shouldn't we align to 4.4.4? (Probably not, just our of curiosity)

argshook:
> @argshook Is 4.0.5 good enough for TS, shouldn't we align to 4.4.4? (Probably not, just our of curiosity)

i agree, eventually we should. however in this PR i'm not changing the version, just moving dependency declaration to `devDependencies`.

probably should do upgrades as a separate PR, cause upgrading might have some implications

PR #1578, opened 2023-04-18
on hover background change on tutorials

saikrishnatbijil:
on hovering the chips in tutorials page the border color only changes, So i have added background color change on hover and when selected it will be back to border color only.

https://user-images.githubusercontent.com/98866798/232779423-aa0a9bce-667a-46f5-aa89-468f56bbe05f.mov
Screen Shot 2023-04-18 at 6 01 51 PM


saikrishnatbijil:
Does this look good to you or do I need to do much better @argshook

bluestreak01:
this is subjective

PR #1579, opened 2023-04-18
Flow chart Image path is wrong I have updated that

chaitanyarai19:
In the questdb.io section inside the docs in the template section the flow chart image path is wrong I have updated that please merge the pull


argshook:
the static URL is correct, as the guidelines are meant to be rendered during development, in local environment. This is stated in the readme.

PR #1581, opened 2023-04-18
rearrange footer links

argshook:


tris0laris (approved):
good first step. maybe we can improve even further, but this partially address a very quick win. lets push!

argshook:
> [...] maybe we can improve even further, but this partially address a very quick win [...]

yep, that's exactly the idea!

PR #1595, opened 2023-04-19
image overflow at swag page

saikrishnatbijil:
what i did is changed the default size of the image to `300px` and at `min-width: 800px` i changed it to `400px`
**before:**
Screen Shot 2023-04-19 at 10 28 39 PM
**after:**
Screen Shot 2023-04-19 at 10 28 57 PM

### DIFF
--- a/src/css/community/page.module.css
+++ b/src/css/community/page.module.css
@@ -11,6 +11,9 @@
   .flex {
     flex-wrap: nowrap;
   }
+  .section_image {
+    max-width: 400px;
+  }
 }
 
 .section__title {
@@ -87,7 +90,7 @@ html[data-theme="light"] .contribution_link img {
 }
 
 .section_image {
-  max-width: 400px;
+  max-width: 300px;
   width: auto;
   height: auto;
   margin: 0 auto;
--- a/src/css/community/page.module.css
+++ b/src/css/community/page.module.css
@@ -11,9 +11,6 @@
   .flex {
     flex-wrap: nowrap;
   }
-  .section_image {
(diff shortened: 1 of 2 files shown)


argshook (changes requested):
unable to merge, because this change affects the layout on wider screens.

saikrishnatbijil:
> unable to merge, because this change affects the layout on wider screens.
i checked it and it looks normal in wider screens 
Screen Shot 2023-04-20 at 4 43 37 PM

argshook:
It looks normal as in it's not broken. But it looks different from our current view. We would prefer fixes for narrow screens to not affect the UI of wider screens.

Hope that makes sense!

saikrishnatbijil:
@argshook i think i fixed it now

argshook:
> @argshook i think i fixed it now

sorry, i still see an issue.
Please compare wide screen on the preview of this PR against current production version:

# production

Screenshot 2023-04-21 at 10 33 04

# this PR

Screenshot 2023-04-21 at 10 33 14

The layout is different: production version has bigger picture, which has impact on the width of text.

It's not a huge change, but if we're doing a fix, it should not have regressions.

saikrishnatbijil:
@argshook i don't know why this is happening but i tested it, Here is my screen while i was testing it,
So for convenience i toggled light mode on the local website, and dark mode in the production 
https://user-images.githubusercontent.com/98866798/233587269-51036dba-0aab-46ab-ae28-15d056b86f39.mov

argshook (approved):

PR #1609, opened 2023-04-24
chore(docs): document casting from string to UUID and Symbols

jerrinot:
document casting from string to UUID and Symbols


glasstiger (approved):

docs/reference/api/ilp/columnset-types.md
```diff
@@ -196,6 +196,62 @@ tracking,obj=VLCC\ STEPHANIE gh="" 2000000000\n
 values into higher resolution column, will cause the entire line to be rejected.
 :::
 
+### Cast to SYMBOL
+String value can be cast to `symbol` type when the destination column exists and is
+of a `symbol` type already. This gives clients an option to populate `symbol` columns
+without knowing the columns are of the `symbol` type.
```
amyshwang, line 202:
```suggestion
String values can be cast to the `symbol` type when the destination column already exists and is
of a `symbol` type. This allows clients to populate `symbol` columns
without knowing the columns are of the `symbol` type. 
```

docs/reference/api/ilp/columnset-types.md
```diff
@@ -196,6 +196,62 @@ tracking,obj=VLCC\ STEPHANIE gh="" 2000000000\n
 values into higher resolution column, will cause the entire line to be rejected.
 :::
 
+### Cast to SYMBOL
+String value can be cast to `symbol` type when the destination column exists and is
+of a `symbol` type already. This gives clients an option to populate `symbol` columns
+without knowing the columns are of the `symbol` type. 
+
+```sql
```
amyshwang, line 204:
```suggestion
```questdb-sql
```

docs/reference/api/ilp/columnset-types.md
```diff
@@ -196,6 +196,62 @@ tracking,obj=VLCC\ STEPHANIE gh="" 2000000000\n
...
+CREATE TABLE trade (
+    ticker SYMBOL,
+    timestamp TIMESTAMP
+) TIMESTAMP(ts) PARTITION BY HOUR;
+```
+Send message including `BTCUSD` as `string` value:
```
amyshwang, line 210:
```suggestion
Send messages including `BTCUSD` as `string`:
```

docs/reference/api/ilp/columnset-types.md
```diff
@@ -196,6 +196,62 @@ tracking,obj=VLCC\ STEPHANIE gh="" 2000000000\n
...
+
+| timestamp                   | ticker |
+|:----------------------------|:-------|
+| 2021-11-29T16:20:21.000000Z | BTCUSD |
+| 2021-12-01T23:53:41.000000Z | BTCUSD |
+
+It's preferred to send `symbol` values directly as `symbol` type as it will
+automatically create `symbol` column if it doesn't exist. When sending `symbol`
+values as `string` type and the column does not exist then it will be created
+as `string` type.
```
amyshwang, line 226:
```suggestion
It's preferred to send `symbol` values directly as `symbol` type because a `symbol` column will be created if it doesn't exist. When sending `symbol`
values as the `string` type and the column does not exist, then it will be created
as the `string` type.
```

docs/reference/api/ilp/columnset-types.md
```diff
@@ -196,6 +196,62 @@ tracking,obj=VLCC\ STEPHANIE gh="" 2000000000\n
...
+
+### Cast to UUID
+String value can be cast to `uuid` type when the destination column exists and is
```
amyshwang, line 229:
```suggestion
String values can be cast to the `uuid` type when the destination column exists and is
```

docs/reference/api/ilp/columnset-types.md
```diff
@@ -196,6 +196,62 @@ tracking,obj=VLCC\ STEPHANIE gh="" 2000000000\n
...
+of a `uuid` type already and the string value is a valid UUID.
+
+```sql
```
amyshwang, line 232:
```suggestion
```questdb-sql
```

amyshwang (commented):
Editorial review - Amy to add a commit later

docs/reference/api/ilp/columnset-types.md
```diff
@@ -192,10 +192,80 @@ tracking,obj=VLCC\ STEPHANIE gh="" 2000000000\n
...
+When sending `symbol` values as the `string` type and the column does not exist,
+then it will be created as the `string` type.
+
+### Cast to UUID
+
+String values can be cast to the `uuid` type when all the following are true:
+
+- the destination column exists
+- the destination column type is `uuid`
+- the `string` values are valid UUID.
```
amyshwang, line 242: We have three conditions but in L267 we only mention the result if the third condition is not met.

amyshwang (commented):
Adding a comment as I don't know if the aspect I mentioned needs to be cover.

docs/reference/api/ilp/columnset-types.md
jerrinot: That's a very good question! 

The lines 240-242 talk about conditions which must be fulfilled to store a string as a UUID. A string from ILP will be stored as a UUID if and only if all three conditions are met. 

The line 267 talks about explicit rejections. What happens if you have a UUID column and an ILP client sends a string which is not a valid UUID? It's rejected.  

So the conditions differ because they describe different situations.

amyshwang: I had this question because I read:

> We recommend sending symbol values directly as the symbol type because it will automatically create a symbol column if it doesn't exist.

I wonder what would happen if:
- the destination column does not exist --> a column will be created as `string`? Should we document this?

- If the destination column type is not `uuid`, I think the conversion depends on the column type so this can be deducted.

jerrinot: * If a target column does not exist and an ILP client sends a string then QuestDB will create a string column. This is true for any type. If a target column does not exist and a client sends it as `long` then QuestDB will create a `long` column, etc.  Do you think we should explicitly document it here? I am not sure. _(I'm honestly _not sure_, it's not meant to be just a polite form of disagreement from my side :))_

* if a destination column does exist and it is not UUID and an ILP client sends a valid UUID string then the processing fails unless the destination column is `String` or `Symbol`. Why? ILP receiver can cast String into 4 other types:
image

but:
- casting UUID string to char fails because UUID length is >1
- casting to geohash fails - as a UUID is not a valid geohash
- we know the target column is not UUID

So we are left with casting just to `symbol` + storing into a `string` column natively.

amyshwang: Thanks! So I think the notes about sending symbol values as symbol type are just a reminder. I have added some general info at the start of the page. I guess there is no harm in mentioning it. I also move one info block to the start of the page as it applies to all.

jerrinot: right, there is one important difference between casting from String to UUID vs Symbol: ILP supports symbols natively. So while you can send symbols as strings you can also send them as native symbols. it has the advantage of being correctly created as `symbol` when a target column does not exist. that's why the note is there. 

On the other hand: ILP does not support UUID natively (yet). So you have no other option but to send UUID as strings.

docs/reference/api/ilp/columnset-types.md
```diff
@@ -192,10 +194,80 @@ tracking,obj=VLCC\ STEPHANIE gh="" 2000000000\n
...
+:::note
+
+Downcast of `geohash` value, which is inserting of lower resolution values into
+higher resolution column, will cause the entire line to be rejected.
+
 :::
 
+### Cast to SYMBOL
+
+These value correspond to the QuestDB type [`symbol`](/docs/concept/symbol/).
```
jerrinot, line 206: I don't understand what is this sentence meant to say:
> These value correspond to the QuestDB type [`symbol`](/docs/concept/symbol/).

amyshwang: Ah, yes:

```suggestion
The symbol values correspond to the QuestDB type [`symbol`](/docs/concept/symbol/).
```

jerrinot (commented):
@amyshwang your changes look good, but as the original author I am not able to approve it. maybe @glasstiger can?

glasstiger (approved):

glasstiger:
> @amyshwang your changes look good, but as the original author I am not able to approve it. maybe @glasstiger can?

approved

PR #1665, opened 2023-05-06
fixed docs screen in mobile

saikrishnatbijil:
i fixed the docs screen by adding a width value to the class .doc__main
**before:**
Screen Shot 2023-05-06 at 1 15 05 PM
**after:**
Screen Shot 2023-05-06 at 1 15 47 PM
not only command-line-options some other docs where also overflowing i fixed that to


saikrishnatbijil:
@argshook could you check on my pr

argshook (approved):
hey @saikrishnatbijil apologies for just now getting to your PR.

we should take a better look at responsive layout of `/docs/*` pages, as setting width is more of a band-aid than solution to the symptom.

but it solves the issue and that's what matters!

thanks!

PR #1680, opened 2023-05-12
feat(docs): adding data durability wording

amyshwang:
Capturing https://github.com/questdb/questdb/pull/3304


docs/concept/storage-model.md
```diff
@@ -73,6 +73,9 @@ synchronous or asynchronous IO. The `msync()` calls are made for column files
 only, so while the `sync`/`async` commit modes improve the overall durability,
 they don't guarantee durability in the face of OS errors or power loss.
 
+From QuestDB 7.2 onwards, `fsync()` calls are used to improve durability over OS
```
puzpuzpuz, line 76: We were using msync/fsync before 7.2. What 7.2 changes is that it adds a bunch of fsync/msync for both metadata and data files. Metadata files are sync always (most, but not all of them if we keep @bluestreak01 's PR as is) while data files are synced only when sync commit mode is configured by the user.

amyshwang: I amended the wording - thank you.

docs/concept/storage-model.md
```diff
@@ -73,6 +73,11 @@ synchronous or asynchronous IO. The `msync()` calls are made for column files
 only, so while the `sync`/`async` commit modes improve the overall durability,
 they don't guarantee durability in the face of OS errors or power loss.
 
+From QuestDB 7.2 onwards, data durability is improved over OS errors or power
```
puzpuzpuz, line 76: I'd suggest removing this paragraph completely and change the previous one in the following way:

By default, QuestDB relies on OS-level data **durability** for data files leaving
the OS to write dirty pages to disk. Data durability can also be configured to invoke
`msync()`/`fsync()` on each commit at the cost of reduced ingestion throughput.
Consider enabling `sync` commit mode to improve data durability in the face of
OS errors or power loss:
```ini title=server.conf
cairo.commit.mode=sync
```

puzpuzpuz: cc @bluestreak01 who might want to change the wording

amyshwang: Updated using the suggestion!

docs/concept/storage-model.md
```diff
@@ -64,14 +64,17 @@ does so lock-free to ensure minimal impact on concurrent reads.
...
-synchronous or asynchronous IO. The `msync()` calls are made for column files
-only, so while the `sync`/`async` commit modes improve the overall durability,
-they don't guarantee durability in the face of OS errors or power loss.
+By default, QuestDB relies on OS-level data **durability** for data files
+leaving the OS to write dirty pages to disk. Data durability can also be
+configured to invoke `msync()`/`fsync()` for column files on each commit at the
+cost of reduced ingestion throughput. Consider enabling the `sync` commit mode
+to improve data durability in the face of OS errors or power loss:
+
+```txt
```
puzpuzpuz, line 75:
```suggestion
```ini title="server.conf"
```

PR #1681, opened 2023-05-16
feat(docs) documenting overview for SQLAlchemy

amyshwang:


sidebars.js
```diff
@@ -131,6 +131,14 @@ module.exports = {
           type: "link",
           href: "/blog/integrate-apache-spark-questdb-time-series-analytics",
         },
+        {
+          label: "SQLAlchemy",
+          customProps: {
+            tag: "New",
+          },
+          type: "link",
+          href: "/docs/third-party-tools/sqlalchemy",
```
argshook, line 140: this code should work but the intended way to link between docs pages is this:

```suggestion
          customProps: {
            tag: "New",
          },
          type: "doc",
          id: "third-party-tools/sqlalchemy",
```

amyshwang: Question - should I add a trailing slash to these, similar to all the hyperlinks we have been amending?

```suggestion
          label: "SQLAlchemy",
          customProps: {
            tag: "New",
          },
          type: "link",
          href: "/third-party-tools/sqlalchemy/",
```

argshook: That's optional, but for consistency it would be nice

docs/third-party-tools/sqlalchemy.md
```diff
@@ -0,0 +1,82 @@
...
+---
+
+[SQLAlchemy](https://www.sqlalchemy.org/) is an open-source SQL toolkit and ORM
+library for Python. It provides a high-level API for communicating with
+relational databases, including schema creation and modification, an SQL
+expression language, and database connection management. The ORM layer abstracts
+away the complexities of the database, allowing developers to work with Python
+objects instead of raw SQL statements.
+
+QuestDB implements a dialect for SQLAlchemy using the
```
javier, line 13: I think we are missing a word here "using the PACKAGE questdb-connect"

docs/third-party-tools/sqlalchemy.md
```diff
@@ -0,0 +1,82 @@
...
+class Signal(Base):
+    __tablename__ = 'signal'
+    __table_args__ = (qdbc.QDBTableEngine('signal', 'ts', qdbc.PartitionBy.HOUR, is_wal=True), )
+    source = Column(qdbc.Symbol)
+    value = Column(qdbc.Double)
+    ts = Column(qdbc.Timestamp, primary_key=True)
+
+
+def main():
+    engine = create_engine('questdb://localhost:8812/main')
```
javier, line 55: I find it confusing that we are not using the complete connection string here. A complete string would be 
`engine = create_engine('questdb://admin:quest@localhost:8812/qdb')`

Otherwise the defaults are used. I also find very confusing we are using `main` as the database name. It seems the db name is ignored, but since default is qdb, I would just go with that one

docs/third-party-tools/sqlalchemy.md
```diff
@@ -0,0 +1,82 @@
...
+            ))
+        ## The insert() SQL expression construct
+
+    finally:
+        if engine:
+            engine.dispose()
+
+
+if __name__ == '__main__':
+    main()
```
javier, line 75: This is not working for me. Unless I add a commit statement I cannot see my values after inserting. Is it working for others?

javier: Also, I think we should probably show the alternative way of inserting/selecting in a non-ORM way. Maybe just adding an INSERT statement with parameters and a select as seen at https://gist.github.com/javier/484de305326a6b77747e6c1e20611dfa

To make it easier we can just do the INSERT and select over the signal table

javier (changes requested):

javier (approved):

javier (approved):

javier (approved):

PR #1683, opened 2023-05-16
feat(docs) adding rank() function

amyshwang:


docs/reference/function/analytic.md
```diff
@@ -18,6 +18,74 @@ Analytic functions are used with an `OVER` clause to define the way data is
 grouped and processed. The `OVER` clause is used with `PARTITION BY` and
 `ORDER BY` to set unique parameters and organize the rows.
 
+## rank
+
+`rank()` - returns the rank of the current row with gaps within the result set.
```
puzpuzpuz, line 23: nit: I'd mention groups and row_number here just like pg docs do it: "Returns the rank of the current row, with gaps; that is, the row_number of the first row in its peer group."

docs/reference/function/analytic.md
```diff
@@ -18,6 +18,74 @@ Analytic functions are used with an `OVER` clause to define the way data is
...
+The following query uses `rank()` to display output based on the rating:
+
+```questdb-sql
+SELECT location, price, date_sold,
+  rank() OVER (ORDER BY rating ASC) rank
+FROM 'housing'
+ORDER BY rank;
+```
+
+| location       | price    | date_sold                   | rank |
```
puzpuzpuz, line 76: nit: would be nice to include the `rating` column here, so that the reader doesn't have to look it up in the above table.

puzpuzpuz (approved):

PR #1689, opened 2023-05-18
feat(blog): TSBS blog post

amyshwang:


blog/2023-05-18-optimizing-optimizer-questdb-time-series-benchmark-suite.mdx
```diff
@@ -0,0 +1,225 @@
...
+<Screenshot
+  alt="A screenshot of the CPU profile graph from the pprof web UI"
+  title="A glimpse of the CPU profile graph from pprof web UI"
+  height={469}
+  src="/img/blog/2023-05-18/cpu-profile.png"
+  width={1056}
+/>
+
+Here, the `NextItem` calls were doing buffered disk I/O, which is inevitable in
+the TSBS case. That's because first, TSBS generates an input file with ILP
```
amyshwang, line 108: Here is the "first" explaining one of the two reasons?

Asking because the reasons that the `NextItem` calls were doing buffered disk I/O, which is inevitable in the TSBS case are:
1. TSBS generates an input file with ILP
messages, and then a separate loader program is used to send the data to the
target database.
2. The `reflect.Select` function is essential for the so-called flow control code
used in the TSBS loader and it's not easy to swap with something faster, so we
decided to look into it later and check the `Append` code first.

In between, we introduce two remaining call sites and this makes it a bit confusing as I got lost between "the second reason" or "the second function".

I suggest:

```
That's because first, TSBS generates an input file with ILP messages, and then a separate loader program is used to send the data to the target database. 

Second, the two remaining significant call sites are the `Append` function from the QuestDB loader and the `reflect.Select` function from the standard library. The `reflect.Select` function is essential for the so-called flow control code used in the TSBS loader. Swapping with something faster is not simple. 

Therefore, we decided to look into it later and check the `Append` code first.
```
Not sure if this is still okay so adding it as a comment.

puzpuzpuz: Here "first" stands for the first of two steps required to run TSBS benchmark: first, you need to generate the data, second, you need to run the loader to send the data to the database. Maybe it deserves rewording. WDYT?

puzpuzpuz (approved):

PR #1690, opened 2023-05-18
chores(docs): improve workflow for Superset

amyshwang:
I was able to start Superset following the new steps. However, when I log in Superset via http://localhost:8088/ + admin sign in, it was linking to my previous trial.
As a result, also database connection tested okay, I was not able to connect to a new questdb instance.
I can also see the previous dataset but I cannot delete it.


docs/third-party-tools/superset.md
```diff
@@ -33,54 +35,63 @@ pip install questdb-connect
 
 Once installed and initialized, Superset is accessible via `localhost:8088`.
 
+Sign in with the following details:
+
+- Username: admin
+- Password: admin
+
```
amyshwang, line 42: I got an invalid decryption key when connecting to the database although the connection test was successful.
I could access the UI seeing my previous dataset but when adding new one following the tutorial, I got "invalid decryption key" error.

javier: This is because your volumnes for these docker containers were previously created and used a different key. 
You can start completely from scratch by executing

```
docker compose down -v
```
That will remove your docker volumes for your superset docker containers. Next time you start volumes will be empty (and initialized by superset with the default admin user). That will also solve the decryption key error

docs/third-party-tools/superset.md
```diff
@@ -7,18 +7,20 @@ description: Guide for using Superset with QuestDB
...
-implements the SQLAlchemy and Superset engine, to integrate
-Superset with QuestDB.
+QuestDB provides the
+[QuestDB Connect](https://pypi.org/project/questdb-connect/) python module that
+implements the SQLAlchemy and Superset engine, to integrate Superset with
+QuestDB.
 
 ## Requirements
 
 - Python from 3.8.x to 3.10.x
```
amyshwang, line 17: Quick note: 
stand alone
pip install qdb connect 
installing superset - https://superset.apache.org/docs/installation/installing-superset-from-scratch

Docker

javier (changes requested):
I am not too happy with the current flow yet. Pushed a commit directly with my proposed changes

docs/third-party-tools/superset.md
```diff
@@ -7,85 +7,83 @@ description: Guide for using Superset with QuestDB
...
-- DISPLAY NAME: QuestDB
-- SQLALCHEMY URI: `questdb://admin:quest@host.docker.internal:8812/main`
+  ```bash
+  git clone https://github.com/apache/superset.git
+  cd superset
+  ```
 
-Once connected, tables in QuestDB will be visible for creating Datasets in
-Superset.
+1. Update the requirement:
```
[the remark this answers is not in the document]
marregui: fixed in a700779988a32092ad17ecfd5bbd2205946cf7d5

marregui (approved):
LGTM I have tested these exact steps, it all works for me.

docs/third-party-tools/superset.md
```diff
@@ -9,73 +9,72 @@ explore data through customizable dashboards and reports.
...
 
-### Installing Superset via QuestDB Connect
+## Installing Superset via QuestDB Connect
 
 If you have a stand-alone installation of Apache Superset and are using Apache
-Superset without Docker, you need to instal QuestDB Connect using `pip`:
+Superset without Docker, you need to install the following requirements :
+
+- Python from 3.9 to 3.11
+- Psycopg2 for connectivity
```
javier, line 63: not needed. Superset will install it as a dependency

docs/third-party-tools/superset.md
```diff
@@ -9,73 +9,72 @@ explore data through customizable dashboards and reports.
...
+- SQLAlchemy for ORM and easy interaction
```
javier, line 64: same as above. Not needed. Superset will install it

javier (approved):
Looking good. Still two requirements that we can remove (from the stand alone section) but already looking good to me

javier (approved):
All good!

PR #1696, opened 2023-05-23
chore(docs): update line.udp.enabled.receiver default

amyshwang:
capturing https://github.com/questdb/questdb/pull/3369/


jerrinot:
should we also add a note here? https://questdb.io/docs/reference/api/ilp/udp-receiver/#configuration

this page is no longer included in the left menu bar, but you can still find it via search: both the builtin search and probably also via Google?

amyshwang:
Yes we did not remove it so it's still searchable. I would not bother adding a line about the default value when we will be phasing this out - is there particular reason why we should consider adding the default value when it is not mentioned in the first place?

jerrinot (dismissed):

jerrinot (approved):

PR #1707, opened 2023-05-24
feat(*): Glossary page

insmac:
This PR adds a new `Glossary` page. 

The list of terms is auto-generated by a `glossary` Docusarusus plugin from the files in `src/pages/glossary`.

Because the content for this page is not yet done, it's currently hidden from the nav menu,


plugins/glossary/index.js
```diff
@@ -0,0 +1,46 @@
...
+const path = require("path");
+
+function readMarkdownFiles(directory) {
+  const files = fs.readdirSync(directory);
+  const glossaryData = [];
+  
+  files.forEach((file) => {
+    const filePath = path.join(directory, file);
+    const fileContent = fs.readFileSync(filePath, "utf-8");
+    const titleMatch = fileContent.match(/^#\s+(.+)$/m);
```
argshook, line 11: maybe can use `import { parseFrontMatter } from "@docusaurus/utils"`

insmac: Yes, we can use the meta tags, sure instead of the title, especially if we want to make the item title a bit different than what's on the list. 🙂 I'll adjust that later, for now, I don't have the content.

argshook (approved):

PR #1713, opened 2023-05-30
amend 3rd party superset to include correct version in Pypi

marregui:


amyshwang:
@marregui could you check if changes in https://github.com/questdb/questdb.io/pull/1690 is a good alternatives? I think we should try to use one PR for this!

marregui:
@amyshwang oh! I did not realise of the existence of that other PR!, I will amend that one instead while I review it :P

marregui:
superseded by https://github.com/questdb/questdb.io/pull/1690

PR #1730, opened 2023-06-05
[DO NOT MERGE] Pagespeed: Disable marquee autoplay on mobile

insmac:
This is an experiment to bring the LCP time (Largest Contentful Paint) time down on mobile - my suspicion is that the marquee's autoplay is to blame, but that's yet to be seen.

On mobile, play state of the marquee can be toggled by tapping anywhere within the logo area.


insmac:
Closing, gathered enough data from subsequent Netlify PR builds and PageSpeed runs.

PR #1731, opened 2023-06-05
feat(glossary): adding data partitioning

amyshwang:
- Adding the entry for data partitioning
- Adding a link to this new page from the "Time partition" concept page


docs/concept/partitions.md
```diff
@@ -55,8 +61,10 @@ We recommend adding time partition to table to benefit from the following advant
 The following SQL keyword and function are implemented to present the partition
 information of a table:
 
-- The SQL keyword [SHOW PARTITIONS](/docs/reference/sql/show/#show-partitions) returns general partition information for the selected table.
-- The function [table_partitions('tableName')](/docs/reference/function/meta/) returns generation partition and allows filtering for the selected table.
+- The SQL keyword [SHOW PARTITIONS](/docs/reference/sql/show/#show-partitions)
+  returns general partition information for the selected table.
+- The function [table_partitions('tableName')](/docs/reference/function/meta/)
+  returns generation partition and allows filtering for the selected table.
```
javier, line 67: I don't like this phrasing "and allows filtering for the selected table". It is confusing and even after executing both show partitions and the table_partitions function it took me a few seconds to understand what's going on. Both SHOW partitions and table_partitions produce the exact same information. The difference is that SHOW can be used only to do just that, while table_partitions is a metadata SQL function that returns the values as a table, so it can be  used from a SELECT statement. That means data can be filtered, but also specific columns can be selected and data can be joined. I would probably rephrase as 'the meta function table_partitions(table_name) returns the same information as SHOW PARTITIONS, but it can be invoked from a SELECT statement, so WHERE, JOINs, UNIONs... are supported'

javier: Related to this, I think the meta page is inconsistent on how it documents the different functions https://questdb.io/docs/reference/function/meta

src/pages/glossary/database-partitioning.md
```diff
@@ -0,0 +1,128 @@
...
+import Screenshot from "@theme/Screenshot"
+
+# What is database partitioning?
+
+Database partitioning (or data partitioning) is a technique used to split data
+in a large database into smaller chunks called partitions. Each partition is
+then stored and accessed separately to improve the performance and scalability
+of the database system.
+
+Database partitioning strategy applies to different types of databases such as
```
javier, line 15: should it be strategy or strategies? This is probably just my poor English skills :)

src/pages/glossary/database-partitioning.md
```diff
@@ -0,0 +1,128 @@
...
+  alt="Diagram showing an example of horizontal partitioning"
+  height={342}
+  src="/img/glossary/data-partitioning/horizontal-partitioning.webp"
+  width={900}
+  title="Horizontal partitioning"
+/>
+
+On the other hand, horizontal partitioning works by splitting the table by rows
+based on the partition key. In this approach, each row of the table is assigned
+to a partition based on some criteria, which include:
```
amyshwang, line 81: Amy to change the list to a bullet list as these are unrelated items.

src/pages/glossary/database-partitioning.md
```diff
@@ -0,0 +1,128 @@
...
+  src="/img/glossary/data-partitioning/vertical-partitioning.webp"
+  width={900}
+  title="Vertical partitioning"
+/>
+
+In vertical partitioning, columns of a table are divided into partitions with
+each partition containing one or more columns from the table. This approach is
+useful when some columns are accessed more frequently than others. Frequently
+accessed columns may be split into different partitions running on discrete
+servers. Alternatively, columns that are rarely used may be partitioned to a
```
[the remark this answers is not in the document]
javier: Yes. Sharding means you divide the whole table intro partitions across multiple servers, but then all the servers have the same schema. Imagine facebook users table, where users living in Europe might be stored at a table in an EU datacenter, while users in America might be stored at a table in a US datacenter. Both tables have the same schemas, but they keep different ranges of users.

A different strategy might be if you have a users table with, for example, user information (name, email, url...) and user config (subscribed_to_newsletter, billing_address...). If you sharded as we saw before that could be an option, but maybe for some reason you prefer to keep all user info together in a single instance because maybe you want to do cross-user selects. Doing sharding means you need to run a distributed select and then combine. So an alternative would be dividing the table users in two parts: one with the generic user data, the other with the config data. 

Vertical partitioning is, like sharding, usually done to outgrow the capacity of a single server, but it complicates your application as you know have the data distributed. 

On the other hand, horizontal partitioning, the one we do, is done to achieve better performance on one server by skipping the amount of data you need to access if you are filtering by the partition columns

src/pages/glossary/database-partitioning.md
```diff
@@ -0,0 +1,128 @@
...
+[Apache Parquet](https://parquet.apache.org/docs/file-format/) support
+partitioning natively, making it ideal for big data processing.
+
+## Partitioning in QuestDB
+
+QuestDB supports
+[partitioning tables](https://questdb.io/docs/concept/partitions/) by intervals
+of time, including hour, day, week, month, and year. Tables that are partitioned
+by time perform better for timestamp interval searches, as QuestDB’s SQL
+optimizer engineer leverage partitioning to reduce disk I/O. QuestDB also
```
javier, line 125: I am not sure this is grammatically solid "as QuestDB’s SQL
optimizer engineer leverage partitioning to reduce disk I/O"

javier (changes requested):
Added my comments

docs/concept/partitions.md
```diff
@@ -34,13 +39,14 @@ import Screenshot from "@theme/Screenshot"
 | --------------- | ---------------- |
 | `HOUR`          | `YYYY-MM-DD-HH`  |
 | `DAY`           | `YYYY-MM-DD`     |
```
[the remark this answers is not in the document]
amyshwang: Yes - I just double checked - https://github.com/questdb/questdb.io/pull/1647/commits/26abf6fac94db0d92e889fa16e2ce40af8497956#r1184771580

src/pages/glossary/database-partitioning.md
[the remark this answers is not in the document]
amyshwang: Okay so vertical partitioning is a way of sharding?
It is confusing to me because we have a section to illustrate the difference between partitioning and sharding and here we seems to contradict ourselves:
The general definition:

>Database partitioning deals with a single database instance, whereas sharding
splits partitions (shards) across multiple database instances for scalability
and availability.

And here:

>Frequently
accessed columns may be split into different partitions running on discrete
servers.

I found this online:
>Often, however, horizontal partitioning splits tables across multiple servers for the purposes of increasing scalability. This is called sharding.

https://www.cockroachlabs.com/blog/what-is-data-partitioning-and-how-to-do-it-right/

I will suggest some changes based on this information.

src/pages/glossary/database-partitioning.md
amyshwang: Updated - good catch - thanks!

src/pages/glossary/database-partitioning.md
amyshwang: Updated the text.

docs/concept/partitions.md
```diff
@@ -22,7 +26,8 @@ import Screenshot from "@theme/Screenshot"
 
 - Partitioning is only possible on tables with a
   [designated timestamp](/docs/concept/designated-timestamp/).
-- Available partition intervals are `NONE`, `YEAR`, `MONTH`, `WEEK`, `DAY`, and `HOUR`.
+- Available partition intervals are `NONE`, `YEAR`, `MONTH`, `WEEK`, `DAY`, and
+  `HOUR`.
 - Default behavior is `PARTITION BY NONE` when using
   [CREATE TABLE](/docs/reference/sql/create-table/) and `PARTITION BY DAY` via
   [ILP ingestion](/docs/reference/api/ilp/overview/).
```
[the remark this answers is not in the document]
amyshwang: Added here - https://github.com/questdb/questdb.io/pull/1733

docs/concept/partitions.md
amyshwang: meta page - To be completed as a separate task

docs/concept/partitions.md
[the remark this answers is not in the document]
javier: thanks!

src/pages/glossary/database-partitioning.md
```diff
@@ -0,0 +1,133 @@
...
+  the function, adding or removing a new partition may require an expensive
+  migration process.
+
+- Composite partitioning: any of the aforementioned methods can be combined. For
+  example, a time-series workload may first be partitioned by time and further
+  split based on another column field.
+
+One thing to note with horizontal partitioning is that performance depends
+heavily on how evenly distributed the data is across the partitions. If the
+database distribution is skewed, the database with the most records will become
```
javier, line 114: Sorry, I think I didn't explain well here. What I meant is to say: "If the
data distribution is skewed, the partition with the most records will become the bottleneck"

javier (changes requested):
One comment needs addressing. All good otherwise

PR #1736, opened 2023-06-07
WIP: updating glossary term titles

amyshwang:
Currently we:
- present an alphabetical list of terms on the glossary landing page 
- each term as listed on the landing page is also the title (for search result)
- h1 = "What is [Term]?" (sentence case)

We would like to:
- present an alphabetical list of terms on the glossary landing page 
- each term should have "What Is [Term]?" as the title (for search result) - this should be in book title case
- h1 is the same as the defined title


amyshwang:
Fixed by https://github.com/questdb/questdb.io/pull/1737

PR #1737, opened 2023-06-07
add support for `index_title` property in glossary pages

argshook:
This PR adds support for `index_title` on `pages/glossary/*.md` files

The questdb.io/glossary index will be generated as follows:

1. If `index_title` is defined, it will be used as the title
2. otherwise, the title will be generated from the filename:
    1. the filename will be split on `-`
    2. first characters of each word will be capitalized

thus, when `pages/glossary/time-series-database.md`:

* has `index_title: Time-Series Database`, the title will be `Time-Series Database`
* does not have `index_title`, the title will be `Time Series Database`


src/pages/glossary/time-series-database.md
```diff
@@ -1,5 +1,6 @@
 ---
-title: Time-Series Database
+index_title: Time-Series Database
+title: What Is a Time-Series Database?
```
amyshwang, line 3: Does this mean that we no longer need to define the h1 title as it should be the same as the `title`?

amyshwang (commented):
I will update the branch and get Database Partitioning fixed in this PR too.

src/pages/glossary/time-series-database.md
argshook: it's possible to do this way, but it's more work, unfortunately.

Perhaps for now we can have both?

amyshwang: No - I removed h1 and the title is rendered on the page..

amyshwang (approved):
Looks good!

PR #1740, opened 2023-06-08
feat(docs): glossary entry for CAP theorem

amyshwang:


src/pages/glossary/cap-theorem.md
```diff
@@ -0,0 +1,58 @@
...
+
+Availability, on the other hand, guarantees that every request will return a non-error response. However, the response may not reflect the state from the most recent write. 
+
+## Partition tolerance
+
+Partition tolerance means that the system as a whole will continue to operate despite network failures. 
+
+## Implications of the CAP theorem
+
+The limitations imposed by the CAP theorem drive the fundamental behaviors of distributed systems in the event of a network partition failure. In a failure scenario, distributed systems can choose either of the following:
```
amyshwang, line 39: "a network partition" should be a network failure??

amyshwang: Confirmed this is correct as it is.

src/pages/glossary/cap-theorem.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+- Be consistent and partition tolerant (CP) to return a consistent response, sacrificing temporary outages.
+- Prioritize availability and partition tolerance (AP) to always respond to requests even if it means returning stale or conflicting responses. 
+
+It is important to note that the CAP theorem does not dictate that a distributed system must always choose CP or AP at all times. The tradeoff only comes into play during a network partition where consistency or availability is favored.
```
jerrinot, line 45: maybe we could add this sentence: _[PACELC theorem](https://en.wikipedia.org/wiki/PACELC_theorem) is an extension of the CAP theorem which describes trade-offs when the system is running normally in the absence of partitions._  wdyt?

jerrinot (dismissed):
looks good to me. feel free to include the recommendation, I thought it might be useful for some of the more curious readers. it's also OK not to include it.

jerrinot (approved):

PR #1743, opened 2023-06-09
feat(docs): adding glossary entry for stream processing

amyshwang:
- Adding new page
- Reviewing third-part tool page to link terms to the glossary
- Reviewing relevant blog posts to add to the glossary page for stream processing


src/pages/glossary/stream-processing.md
```diff
@@ -0,0 +1,62 @@
...
+Stream processing, or complex event processing (CEP), is a data processing technique that collects, transforms, and analyzes streams of data in real time. Stream processing systems are designed to handle large amounts of high-velocity events (e.g., IoT sensor readings, financial market data, server usage metrics) and provide real-time insights. 
+
+## Stream vs. batch processing
+
+Unlike traditional data pipelines that use batch processing where data is analyzed in batches, stream processing analyzes data on the fly. Since analysis occurs as soon as data is ingested, stream processing systems enable use cases where real-time decision-making is useful. Examples include anomaly detection, fraud prevention, and predictive analytics. 
+
+<Screenshot
+  alt="Batch processing visualization"
+  height={342}
+  src="/img/glossary/stream-processing/batch-processing.webp"
```
jerrinot, line 20: this picture does not work for me:(
image

I am having a hard time interpreting it. 
An idea: Could you try to visualize an "overnight batch processing"? overnight batch processing is/was a common strategy.

src/pages/glossary/stream-processing.md
```diff
@@ -0,0 +1,64 @@
...
+  width={770}
+  title="Batch processing collects events and processes them in batch."
```
jerrinot, line 30: perhaps we could add: _This incurs latency penalty due to the lag between "event accumulated in a batch" and "the batch was actually processed"_. 

Or perhaps this is too long for an image description.  but I feel we should explicitly say the thing about latency in batch processing. as that's *the* reason to use stream processing instead of batches.

jerrinot (approved):

PR #1746, opened 2023-06-12
feat(docs): capture comparison operators

amyshwang:


docs/reference/operators/comparison.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+## > (greater than) and < (less than)
+
+- `(value1) > (value2)` - returns true if `value1` is greater than `value2`.
+- `(value1) < (value2)` - returns true if `value1` is less than `value2`.
+
+### Arguments
+
+- `value1` is any numeric or `string` value.
```
puzpuzpuz: `char` values are also supported.

docs/reference/operators/comparison.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Return value
+
+Return value type is boolean.
+
+### Examples
+
+```questdb-sql
+
+SELECT '5' = '5';
```
puzpuzpuz, line 32: nit: all examples currently use chars, not strings. By default, '5' literal is considered to be a char. You can make it a proper string with `'5'::string` or by using a multi-char literal like `'123'`.

amyshwang: I added these:
```
SELECT 'abc' > 'def';
-- Returns false

SELECT '123' < '456';
-- Returns true
```

bziobrowski (approved):
👍

PR #1756, opened 2023-06-20
feat: New enterprise page

insmac:
WIP


src/components/FeatureTable/styles.module.css
```diff
@@ -104,6 +103,77 @@ html[data-theme='light'] .root tbody tr:nth-child(2n+1) {
   cursor: default;
```
argshook, line 103: i would remove this, link should have a pointer cursor, not default.

src/components/FeatureTable/styles.module.css
```diff
@@ -104,6 +103,77 @@ html[data-theme='light'] .root tbody tr:nth-child(2n+1) {
...
+.rowWithDescriptions {
+  cursor: pointer;
+}
+
+.rowWithDescriptions:hover {
+  background-color: #2F313C;
+}
+
+html[data-theme="light"] .rowWithDescriptions:hover {
+  background-color: #dee0e8 !important;
```
argshook, line 144: maybe it's possible to avoid the `!important`s in this file?

src/modules/cloud/CompareFeatures/index.tsx
```diff
@@ -1,165 +1,264 @@
...
+    key: "cloud",
+  },
+]
+
+const coreFeaturesRows: Row[] = [
   {
     title: "High-throughput ingestion",
-    inOpenSource: "available",
-    inCloud: "available",
+    description: "",
```
argshook, line 21: `description` seems empty everywhere, maybe it's not needed?

src/components/FeatureTable/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+            <th key={col.key}>{col.title}</th>
+          ))}
+        </tr>
+      </thead>
+
+      <tbody>
+        {rows.map((row, index) => (
+          <React.Fragment key={row.title}>
+            <tr
+              {...(withDescriptions === true &&
```
argshook, line 73:
```suggestion
              {...(withDescriptions &&
```

src/components/FeatureTable/styles.module.css
insmac: sure, that's most likely a bug, wrong value from an autocomplete. good catch

src/modules/cloud/CompareFeatures/index.tsx
insmac: good catch, indeed it is an optional prop, which has been introduced in a later commit. Will clean it up.

src/components/FeatureTable/index.tsx
insmac: That would be my initial approach as well, but this will yield a typescript error because of our tsconfig. We can change it though, if we want

src/components/FeatureTable/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+                row.description && {
+                  onClick: () =>
+                    setActiveRow(activeRow === index ? null : index),
+                })}
+              className={clsx({
+                [style.rowWithDescriptions]:
+                  withDescriptions === true && row.description,
```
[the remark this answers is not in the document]
insmac: See comment above

src/components/FeatureTable/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+                [style.activeRow]: activeRow === index,
+              })}
+            >
+              <td className={style.rowTitle}>
+                {withDescriptions === true && (
```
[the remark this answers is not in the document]
insmac: likewise

src/components/FeatureTable/index.tsx
argshook: ah, right, it's because `withDescriptions` is marked as optional on line [19](https://github.com/questdb/questdb.io/blob/ca129e05b0a4b7296c5641f62d7f83ebd0a8e5dd/src/components/FeatureTable/index.tsx#L19).

since it's optional, you could add a default value on line [54](https://github.com/questdb/questdb.io/blob/ca129e05b0a4b7296c5641f62d7f83ebd0a8e5dd/src/components/FeatureTable/index.tsx#L54): `  withDescriptions = false,
`

argshook (approved):

PR #1757, opened 2023-06-23
docs: fix postgres datasource username

gabor-boros:
As per the [documentation](https://questdb.io/docs/third-party-tools/grafana/#add-a-data-source) the username for the datasource should be `admin` instead of `user`.


amyshwang:
Hi @gabor-boros thank you - I dug into this and realised that the one on the docs is wrong - in `server.conf` the configuration is:
```
#pg.readonly.password=quest
#pg.readonly.user=user
```
So the default should be `user`. This is explained in the blog but I should make it clearer.
I will close this PR and submit another one to fix the docs - thank you again for identifying this.

gabor-boros:
@amyshwang Is the wrong config included in the already published Docker images too? Asking this because I was able to connect Grafana to QuestDB using the docs, but not the tutorial. I.e. using admin:quest

gabor-boros:
@amyshwang I'm not sure this is the relevant `server.conf`, but this sats `admin` for the user: https://github.com/questdb/questdb/blob/master/core/src/main/resources/io/questdb/site/conf/server.conf#L604

amyshwang:
@gabor-boros I will find out more about `pg.user=admin` but the one the blog and our latest docs refer to is `pg.readonly.user=user`.

See https://questdb.io/blog/time-series-monitoring-dashboard-grafana-questdb/

gabor-boros:
@amyshwang Ah, this was what I'm missing: `QDB_PG_READONLY_USER_ENABLED=true`. Otherwise, the readonly user is not available for login.

PR #1759, opened 2023-06-26
chore(glossary): change time series data to time-series data

amyshwang:
Search and replace in the time-series database glossary entry page.


argshook (dismissed):

amyshwang:
This is a very simple fix so if we're not merging it now, I think we could close it and open a new one when needed.

PR #1781, opened 2023-07-05
feat(glossary): object storage

amyshwang:


src/pages/glossary/object-storage.md
```diff
@@ -0,0 +1,69 @@
...
+be stored at top-level or nested under other folders. To retrieve a specific
+file, the path to that file must be known. Multiple devices (e.g., VMs, servers)
+can access file storage concurrently. Compared to object storage systems, file
+storage systems are less scalable, but can be a good option if a hierarchical
+structure is already in place or if support for POSIX operations (e.g., updates,
+random reads) or standard file systems are required.
+
+### Block storage
+
+Block storage systems break files into discrete chunks called blocks and store
```
javier, line 43: I think we should add a sentence before the first one. Something like "From a user's perspective, a block storage system looks and behaves like a locally mounted disk drive. However, behind the scenes, block storage systems break files (...) "

src/pages/glossary/object-storage.md
```diff
@@ -0,0 +1,69 @@
...
+them separately. Each chunk of data is assigned a unique identifier, which is
+used to retrieve and assemble into the full file when needed. Block storage is
+best suited for transactional workloads or use cases that require consistent
+performance or lowest latency as block storage is treated as a local disk . In
+general, block storage is often cheaper than file storage systems, but has lower
+availability as only a single device can attach to it. There are some cloud
+offerings that allow multiple devices to attach, but they are limited for
```
javier, line 50: not sure if we should say "they are limited TO" rather than "they are limited FOR"

javier (commented):
Added a couple of small remarks that I skipped when originally reviewing on slab. Sorry!

src/pages/glossary/object-storage.md
amyshwang: Changed to 'to'.

src/pages/glossary/object-storage.md
```diff
@@ -0,0 +1,70 @@
...
+file, the path to that file must be known. Multiple devices (e.g., VMs, servers)
+can access file storage concurrently. Compared to object storage systems, file
+storage systems are less scalable, but can be a good option if a hierarchical
+structure is already in place or if support for POSIX operations (e.g., updates,
+random reads) or standard file systems are required.
+
+### Block storage
+
+From a user's perspective, block storage systems look and behave like a locally
+mounted disk drive. However, behind the scene, block storage systems break files
```
javier, line 44: scenes plural?

javier (approved):

src/pages/glossary/object-storage.md
amyshwang:
```suggestion
mounted disk drive. However, behind the scenes, block storage systems break files
```

javier (approved):

PR #1791, opened 2023-07-10
chore(): display `IoT` tag instead of `IOT`

argshook:
We want `IoT` tag on https://questdb.io/blog to be displayed as `IoT` and not `IOT`. The changes in this PR achieves this.

image

Commit log:

* remove `isTruncated` components
    we do not display truncated blog posts anymore. Thus, this code is not
    needed
* remove default `text-transform: uppercase` from `Chip` component
* manually uppercase tags
    except for `IoT`


amyshwang (approved):
Checked the output and it looks good.

PR #1792, opened 2023-07-10
chore(blog): adding pandas to the blog landing page as a tag

amyshwang:


src/theme/BlogListPage/index.tsx
```diff
@@ -56,6 +56,7 @@ const prioritizedTags: ChipProps["items"] = [
   "engineering",
   "prometheus",
   { label: "k8s", tag: "kubernetes" },
+  "pandas",
```
amyshwang, line 59: Following the fix in https://github.com/questdb/questdb.io/pull/1791,
Maybe this should be:
```suggestion
  "PANDAS",
```

tris0laris (approved):

PR #1802, opened 2023-07-14
feat(pages): Redesign pricing page

oliver-daniel:
Feature branch documenting changes to the pricing page, primarily surrounding the movement of focus from different cloud pricing plans to three streams of QuestDB products.


src/modules/cloud/ContactForm/index.tsx
```diff
@@ -6,13 +6,13 @@ import emailPattern from "../../../utils/emailPattern"
...
 import clsx from "clsx"
-import type { PricingPlan } from "../../pricing/plan"
+import type { ProductInfo } from "../../pricing/product"
 import Link from "@docusaurus/Link"
 import CameraIcon from "../../../assets/img/camera.svg"
 import SvgImage from "../../../components/SvgImage"
 
 type Props = {
-  interestedIn?: PricingPlan["type"] | "custom" | "cloud" | "sla"
+  interestedIn?: ProductInfo["type"] | "custom" | "cloud" | "sla"
```
argshook, line 15: need to make sure that contact forms still need this `interestedIn` value

src/modules/cloud/ContactForm/index.tsx
```diff
@@ -6,13 +6,18 @@ import emailPattern from "../../../utils/emailPattern"
...
 import style from "./styles.module.css"
 import clsx from "clsx"
-import type { PricingPlan } from "../../pricing/plan"
 import Link from "@docusaurus/Link"
 import CameraIcon from "../../../assets/img/camera.svg"
 import SvgImage from "../../../components/SvgImage"
 
 type Props = {
-  interestedIn?: PricingPlan["type"] | "custom" | "cloud" | "sla"
+  interestedIn?:
```
insmac, line 14: Feel free to remove `interestedIn` altogether. It's not used anymore, as the AWS API route used only covers the Cloud CTA.

argshook:
hey @oliver-daniel, i opened a PR to your PR with some suggested changes. Tried to make the code a bit more simple where possible.

https://github.com/questdb/questdb.io/pull/1804

src/components/Slider/index.tsx
```diff
@@ -0,0 +1,44 @@
+import React from "react"
+import * as RadixSlider from "@radix-ui/react-slider"
+import style from "./style.module.css"
+
+export type Props = {
+  // name: string
```
argshook, line 6:
```suggestion
```

src/components/Slider/style.module.css
```diff
@@ -0,0 +1,95 @@
...
+
+.range {
+  position: absolute;
+  background-color: var(--palette-pink);
+  border-radius: 1rem;
+  height: 100%;
+}
+
+.thumb {
+  --thumb-size: 1.8rem;
```
argshook, line 42:
```suggestion
  --thumb-size: 1.5rem;
```

1.8 is a bit too much IMO

src/components/Slider/style.module.css
```diff
@@ -0,0 +1,95 @@
...
+.thumb::after {
+  display: flex;
+  content: attr(data-label);
+  position: absolute;
+  white-space: nowrap;
+  top: calc(100% + var(--tooltip-gap));
+  left: 50%;
+  transform: translateX(-50%);
+  border-radius: 3px;
+  padding: 1rem;
```
argshook, line 69:
```suggestion
  padding: 0.3rem 0.5rem;
```

src/components/Toggle/index.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react"
+import style from "./style.module.css"
+import clsx from "clsx"
+
+type Labellable = string | { label: string }
```
argshook, line 5: is there a good reason to have union type? maybe it's enough to support just `string`?

src/modules/pricing/pricing-calculator/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+      disabled: !available,
+    }))}
+  />
+)
+
+type RateUnit = "hour" | "month"
+
+export const PlanCalculator = () => {
+  const [region, setRegion] = useState<RegionKey>("us-east-2")
+  // const [provider /*, setProvider */] = useState<ProviderKey>("aws")
```
argshook, line 39:
```suggestion
```

src/modules/pricing/pricing-calculator/prices.ts
```diff
@@ -0,0 +1,210 @@
...
+export type CPUSpec = {
+  cores: number
+  ramGB: number
+}
+
+export type StorageSpec = {
+  storageGB: number
+}
+
+// the actual possible regions
```
argshook, line 23:
```suggestion
```

src/modules/pricing/pricing-calculator/prices.ts
```diff
@@ -0,0 +1,210 @@
...
+export const regions = [
+  { key: "us-east-2", label: "Ohio", available: true },
+  { key: "us-west-2", label: "Oregon", available: true },
+  { key: "eu-west-1", label: "Ireland", available: true },
+  { key: "ap-south-1", label: "Mumbai", available: false },
+  { key: "eu-central-1", label: "Frankfurt", available: false },
+] as const
+
+type AvailableRegion = Extract<typeof regions[number], { available: true }>
```
argshook, line 32: is it useful to have a type with only available regions?
maybe there can be just `Regions` and runtime (not typescript type check) decides whether they're available?

src/modules/pricing/pricing-calculator/prices.ts
```diff
@@ -0,0 +1,210 @@
...
+}
+
+export type CPUPricing = CPUSpec & RegionPricing
+//
+export type StoragePricing = StorageSpec & RegionPricing
+
+export const cpuPrices: CPUPricing[] = [
+  { cores: 2, ramGB: 8, region: "us-east-2", price: 0.322 },
+  { cores: 2, ramGB: 8, region: "us-west-2", price: 0.322 },
+  // _0 to preserve trailing zeroes
```
argshook, line 46:
```suggestion
```

IMO this should be a job for the view. The numbers here should not include information needed for view. Maybe [Number.prototype.toFixed() - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) can help here

src/modules/pricing/pricing-calculator/prices.ts
```diff
@@ -0,0 +1,210 @@
...
+  const storage = storagePrices.filter((spec) => spec.region === region)
+
+  return {
+    ...acc,
+    [region]: {
+      cpu,
+      storage,
+    },
+  }
+}, {} as {})
```
argshook, line 210: the `as` casting should be used sparsely, only when really really needed. It's ok now, don't sweat this line, we'll figure it out later on.

src/modules/pricing/pricing-calculator/style.module.css
```diff
@@ -0,0 +1,105 @@
...
+.row {
+  display: flex;
+  flex-direction: row;
+  gap: var(--gap);
+
+  justify-content: space-around;
+  align-items: center;
+
+  flex-wrap: wrap;
+  width: 100%;
```
argshook, line 26:
```suggestion
```

maybe not needed?

src/modules/pricing/products/products.ts
```diff
@@ -0,0 +1,79 @@
...
+    highlight: "$200 credit for new users",
+    ctaLabel: "Get started",
+    pricingLabel: [
+      "from",
+      {
+        // hard-coded from hourly price of
+        // us-east-2 instance with
+        // * 2 CPU
+        // * 8 GB RAM
+        // * 25 GB storage
```
argshook, line 46: useful comment :+1:

src/modules/pricing/pricing-calculator/style.module.css
```diff
@@ -0,0 +1,105 @@
...
+@media only screen and (min-width: 600px) {
+  .card {
+    padding: 2rem 4rem;
+  }
+  .card > .row {
+    flex-basis: 50%;
+  }
+
+  .compressionInfo {
+    width: 50ch;
```
argshook, line 92:
```suggestion
    width: 50ch;
```

maybe this can be expressed via percentage?

argshook (changes requested):

src/components/Toggle/index.tsx
oliver-daniel: Fair point; no guarantee that an object type with a `label` prop would even satisfy the `selected === option` predicate on L27.

src/modules/pricing/pricing-calculator/index.tsx
oliver-daniel: Do you recommend leaving a more "English" comment for posterity (i.e., asserting that the `region` value is only fixed as `aws` until the other providers are implemented), or just trust future programmers' common sense?

src/modules/pricing/pricing-calculator/prices.ts
oliver-daniel: This was more useful in an older version of the PR, where the `pricesByRegion` object would more aggressively assert that it contained pricings for all available regions. Doing it at runtime sounds fine.

src/modules/pricing/pricing-calculator/prices.ts
oliver-daniel: Fair point; I was looking at this from a purely DX perspective, to make it explicit that yes, the thousands place of this price *is* a zero and not an accidentally deleted digit. Prettier wasn't too happy about that, so this was a clunky workaround. Will remove.

src/modules/pricing/pricing-calculator/index.tsx
argshook: Let's implement it when/if needed. Until then a comment can be considered noise. 

https://en.m.wikipedia.org/wiki/You_aren%27t_gonna_need_it

src/components/Toggle/index.tsx
oliver-daniel: The main point was to have a distinction between latent values and user-friendly strings (think country codes vs names), but that's a future problem.

src/modules/pricing/pricing-calculator/style.module.css
oliver-daniel: The main issue I'm encountering is that when the width is anything other than a fixed constant, there's a weird interplay between the length of this content and the overall width of the card as sliders move, resulting in a really disorienting jitter effect. Fixes are very welcomed if you have ideas.

src/components/Toggle/index.tsx
argshook: `Toggle` as a reusable component just gets a list of `options` and calls `onClick` with whichever option was clicked. It doesn't need to know if the option is country name or code or something else.

p.s. `onClick` should really be `onChange`

argshook (dismissed):

argshook (approved):

PR #1804, opened 2023-07-18
chore(): refactor pricing page redesign in #1802

argshook:
- refactor pricing page redesign
- adjust for mobile


src/modules/pricing/products/products.ts
```diff
@@ -33,31 +22,31 @@ export const products: ProductInfo[] = [
...
     title: "Cloud",
     subtitle: "Fully hosted by QuestDB",
-    highlightText: "$200 credit for new users",
+    highlight: "$200 credit for new users",
+    ctaLabel: "Get started",
     pricingLabel: [
       "from",
       {
-        text: formatPrice(cloudPrice, { minimumFractionDigits: 3 }),
+        text: "$0.345",
```
argshook, line 42: pricing is no longer dynamic in this view, so i'm just hardcoding, no need to calc anything

src/modules/pricing/products/products.ts
```diff
@@ -1,30 +1,19 @@
-import { formatPrice } from "../../../utils"
-import { regions } from "../pricing-table/specs"
+import customFields from "../../../config/customFields"
```
argshook, line 1: this file contains a bunch of reusable links. It is used in typescript code (like in this file) and also in `.mdx` files via, for example, such interpolation: `{@slackUrl@}`

src/modules/pricing/products/products.ts
```diff
@@ -1,30 +1,19 @@
...
+export type Label = string | { text: string; emphasized: true }
 
 export type ProductInfo = {
   type: "open-source" | "cloud" | "enterprise"
   title: string
   subtitle: string
-  highlightText?: string
-  pricingLabel: PricingLabel[]
-  pricingSublabel: PricingLabel[]
+  highlight?: string
```
argshook, line 9: while reading this code i thought that type `string` is enough to indicate that value is text, so renamed `highlightText` -> `highlight`.

thanks to typescript my IDE renamed all uses of `highlightText`.

src/modules/pricing/products/products.ts
```diff
@@ -1,30 +1,19 @@
...
+  pricingLabel: Label[]
```
argshook, line 10: IMO the context of this file is enough to know that `Label` is about pricing, so i renamed this type.

src/modules/pricing/products/products.ts
```diff
@@ -1,30 +1,19 @@
...
+  pricingSublabel: Label[]
   specs: string[]
   url: string
-  ctaLabel?: string
+  ctaLabel: string
```
argshook, line 14: the stricter the better!

src/modules/pricing/products/products.ts
```diff
@@ -67,20 +56,17 @@ export const products: ProductInfo[] = [
...
     pricingLabel: [{ text: "Contact us", emphasized: true }],
     pricingSublabel: ["for a quote"],
-    specs: [
-      "Security and authorization solutions",
-      "High availability",
-      "Enterprise SLAs",
-    ],
-    url: "https://questdb.io/enterprise/",
+    specs: ["Security and permissions", "High availability", "Enterprise SLAs"],
+    url: "/enterprise/",
```
argshook, line 69: for internal links it is enough to use relative URL

src/modules/pricing/products/styles.module.css
```diff
@@ -1,44 +1,22 @@
-.title {
-  font-size: 1.3rem;
-}
-
-.products {
-  --color-subtext: rgba(217, 217, 217, 0.51);
+.root {
```
argshook, line 1: using `.root` for top-most component of the file is just a personal taste.

when/if component name changes (or it is refactored/moved to another file), then we don't need to change class name, as `root` is generic.

src/modules/pricing/products/product-card/index.tsx
```diff
@@ -1,71 +1,72 @@
...
     <div className={style.cta}>
-      <Button
-        size="small"
-        dataHook={`get-access-button-product-${product.type}`}
+      <Link
         to={product.url}
-        newTab={false}
+        className={clsx(
+          buttonStyle.button,
+          buttonStyle["button--primary"],
```
oliver-daniel, line 64: By changing the `<Button>` component to a `<Link>`, I see that we now have to import some styling modules from the `Button`'s directory, including what looks like some BEM-style classnames. Can you tell me more about this?

src/modules/pricing/products/products.ts
oliver-daniel: Good to know, thanks!

oliver-daniel (commented):
Thanks for all the comments. I've merged in the majority of the code changes into the branch, with a handful of modifications of my own (mostly comments), that will make this PR appear much smaller.

src/modules/pricing/products/product-card/index.tsx
```diff
@@ -1,71 +1,72 @@
...
+      {product.specs.map((spec, index) => (
+        <li key={index}>{spec}</li>
+      ))}
+    </ul>
+
     <div className={style.cta}>
-      <Button
-        size="small"
-        dataHook={`get-access-button-product-${product.type}`}
+      <Link
```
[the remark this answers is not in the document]
oliver-daniel: To this end, it seems like changes might need to be made at another time to the `Button` module itself, especially lines 75-77 which currently renders `<Button>`s with a non-empty `to` prop as `<a>`.

PR #1806, opened 2023-07-19
[WIP] feat(): Add new pricing calculator panel to pricing page

oliver-daniel:
Tracking changes to this.

As of Jul 18 the presentational components look like crap and a lot of files are in what I think are in the wrong place, but types have been moved around slightly and should be relatively futureproof.


src/modules/pricing/plan-calculator/prices.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const regions = [
+  { key: "us-east-2", label: "Ohio", available: true },
+  { key: "us-west-2", label: "Oregon", available: true },
+  { key: "eu-west-1", label: "Ireland", available: true },
+  { key: "ap-south-1", label: "Mumbai", available: false },
+  { key: "eu-central-1", label: "Frankfurt", available: false },
+] as const
+
+type AvailableRegion = Extract<typeof regions[number], { available: true }>
```
oliver-daniel, line 29: Using this type for the `pricesByRegion` object's keys guarantees that if another region is made available, TS will yell at you to ensure that it is added as a key.

src/modules/pricing/plan-calculator/prices.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const pricesByRegion: { [key in AvailableRegion["key"]]: Spec[] } = {
```
argshook, line 31: @oliver-daniel 
There should not be a need to hardcode `pricesByRegion`

see these, you should be able to reuse:

1. https://github.com/questdb/questdb.io/blob/b5226d71b1efdabf5b79fd73677395b670cacf47/src/modules/pricing/plans/plans.ts#L6-L9
2. https://github.com/questdb/questdb.io/blob/b5226d71b1efdabf5b79fd73677395b670cacf47/src/modules/pricing/pricing-table/specs.ts#L30-L181

oliver-daniel: Agreed; indeed it's the same data reformatted, albeit opinionatedly. I'm kind of approaching it from a "code-as-content" perspective, trying to keep all the "arbitrary"/non-computed information in one place whenever possible, almost like it would appear in the CMS.

insmac:
I'll review this when it's more polished (just because there will be a lot of changes incoming I presume). 
I like using Radix primitive for the slider, as it's headless and we have experience with its implementation.

src/modules/pricing/plan-calculator/prices.ts
argshook: That's fine, you can change the format if it works better in your case.
Just be sure to not duplicate it: i.e. when new format is added, most likely the previous one should be removed.

If both are needed, then let's instead come up with some helper function
which would convert. Happy to help if needed, let me know.

PR #1810, opened 2023-07-24
feat(blog): Blog polish proposal

goodroot:
This PR removes functionality from the blog and makes minor changes for readability.

The thesis is: `less clutter + more readable = more focus/success on a specific/useful action`

Fuzzy KPI: increased time-on-page, decreased bounce and improved engagement within a future single CTA.

--- 

🔮 **For clarity** 

In the below images, note 3 in particular. 

The end is where we want to provide the most valuable & relevant choice to the reader.

Currently, there are 19 options, not counting the floating nav & footer.

This PR reduces down to 8, with potential to refine further.

To this end, we remove:

* Pagination
* Sidebar

👀 **For readability**  

The stylistic changes are:

* Larger text, headers and author subtitle
* Horizontal rule, with support for hr in light mode
* Nudges column over by one increment and shrinks column by one increment

--- 

**Why remove pagination?**

Analytics, trends, and on and on, indicate people find an article via google, social or receive it via a direct link. Very few people read the blog linearly, especially when it spans so many topics and does not provide a cohesive narrative. Docusaurus uses the more doc next/previous primitive for blog, too. It doesn't seem to fit. 

**Why remove the sidebar?**

For similar reasons. It is nice seeing things like version updates in the sidebar. But for really important stuff, there's always top-of-page banners and similar. As it is, it provides a lot of clutter and makes the reading experience busy. It's also linear and doesn't associate content intelligently.

**What's next?**

A follow-up PR can:

1. Redesign current split CTAs (community & letter) and consolidate into a single CTA affordance
2. Move tags above the article, near author section

---

| Before | After |
| --- | --- |
| Screenshot 2023-07-24 at 12 04 31 PM | Screenshot 2023-07-24 at 12 00 23 PM |
| Screenshot 2023-07-24 at 12 05 17 PM | Screenshot 2023-07-24 at 12 02 16 PM |
| Screenshot 2023-07-24 at 12 05 50 PM | Screenshot 2023-07-24 at 12 03 19 PM |


src/theme/BlogPostPage/styles.module.css
```diff
@@ -1,3 +1,11 @@
+.header,
+.markdown,
+.footer {
+  max-width: 665px;
```
argshook, line 4: @goodroot after merging #1813 i tried to keep the same width as i found in your PR. But now i'm thinking maybe it's a bit too narrow? wdyt?

```suggestion
  max-width: 700px;
```

src/theme/BlogPostPage/styles.module.css
```diff
@@ -34,14 +47,26 @@
...
+  font-size: 1.2rem;
+}
+
+.markdown h2,
+.markdown h3,
+.markdown h4,
+.markdown h5,
+.markdown h6 {
+  font-size: 2.2rem;
+}
```
argshook, line 68: we probably should set slightly different font sizes for each heading? Otherwise things might look a bit strange, wdyt?

see example here: https://deploy-preview-1810--unruffled-blackwell-31bfb2.netlify.app/blog/max-open-file-limit-macos-jvm/

few headings of same size are next to each other

src/theme/BlogPostPage/styles.module.css
goodroot: Agree!

src/theme/BlogPostPage/styles.module.css
goodroot: Is now:

Screenshot 2023-07-25 at 11 01 23 AM

argshook (approved):

PR #1816, opened 2023-07-26
Adjust header-blog post gap

argshook:
Increase gap between header and blog post content

Commit log:

* increase gap between header and blog post content


goodroot (approved):
More comfy 👍

PR #1823, opened 2023-07-27
chore(pages): Change 'Need more?' CTA from link to trigger for contact modal

oliver-daniel:
As suggested [here](https://questdb-tech.slack.com/archives/C03FAL4Q52L/p1690479623280299?thread_ts=1690478632.359799&cid=C03FAL4Q52L), changes the smaller CTA in the price calculator on the pricing page to a trigger for a contact dialog.

That being said, this also makes two CTAS in very close proximity open the exact same dialog – this and the "Contact us" button just below – which speaks to me that some better understanding may be necessary to see if this CTA is redundant, should point somewhere else currently on the site, or even requires some new content.


goodroot (approved):
LGTM!

PR #1824, opened 2023-07-27
feat(docs): Proposed fix for preventing 'whiteout' of screenshots with transparency

oliver-daniel:
Aims to fix visibility issues where `Screenshot` images with transparency can have white text/elements be invisible on the white background of the light theme mode.

In particular, this fix augments the `Screenshot` component to allow for the "forcing" of a particular theme mode, overriding the changes to `--ifm-`-prefixed variables to preserve a background of a given colour.

I imagine that there are multiple ways to resolve this problem, ranging from forbidding image types with transparencies (a patently bad idea, imo) to allowing for "alternate" images on varying theme modes (also a bad idea, especially if the secondary image has to be re-created post-hoc and isn't pixel-perfectly identical). This seems like a relatively non-invasive way to handle it; thoughts?


argshook:
> I imagine that there are multiple ways to resolve this problem

absolutely! however, your suggestion works well and is easy to control. I think there's no need to over-engineer here, this PR is good!

argshook (approved):

PR #1833, opened 2023-08-01
Remove commit lag doc & references

goodroot:
Removes initial doc & alters referring documents.

Likely for a second PR (scope this one to commit lag), but ilp (https://questdb.io/docs/reference/api/ilp/tcp-receiver#commit-strategy/) is often mentioned around commit lag.


netlify.toml
```diff
@@ -194,8 +194,8 @@
   to = "/docs/reference/web-console/"
 
 [[redirects]]
-  from = "/docs/guides/hysteresis"
-  to = "/docs/guides/out-of-order-commit-lag/"
+  from = "/docs/guides/out-of-order-commit-lag/"
```
argshook, line 197: what is the purpose of this redirect?

if this redirect exists, then users will not be able to reach `/docs/guides/out-of-order-commit-lag` at all. If that's the intent, then [this file](https://github.com/questdb/questdb.io/pull/1833/files#diff-1d16e670bb5a480b6fdf3c9ad8c83b13a900e59b986f79f323f57b8249295975) should be removed completely

argshook (changes requested):

PR #1840, opened 2023-08-03
chore(): fix layout of author image and title

argshook:
Fix prettier warnings and fix author image and title layout.

Commit log:

* fix prettier warnings
* fix author image and title layout


goodroot (approved):
LGTM -- thx for fixing prettier too. 

Tucked it in a doc revision PR, but this is faster. 👯

PR #1850, opened 2023-08-10
feat(docs): RBAC/ACL tutorial

goodroot:
* adds rbac tut: `/blog/questdb-enterprise-rbac-acls/`


blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+import Screenshot from "@theme/Screenshot"
+
+<Screenshot
+  alt="Three rows of ten diverse lego people - their colourful outfits demonstrate various roles and jobs, and a radical amount of personality."
+  title="Bring the whole gang!"
+  src="/img/blog/2023-08-15/roles.webp"
+/>
+
+Role-based Access Control (RBAC) is a common request from teams with fast growing datasets. **QuestDB Enterprise** and **QuestDB Cloud** deployments now provide a robust, SQL-based ACL syntax to keep permissions organized and your team free from the looming stress of accidental destruction.
```
glasstiger, line 20: suggestion: SQL-based ACL syntax -> SQL-based access control syntax

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+This tutorial demonstrates RBAC basics with both PGWire and Influx Line Protocol (ILP) ingest examples. After reading, you will be ready to administer an instance of any size.
```
glasstiger, line 22: small suggestion: both PGWire and Influx Line Protocol (ILP) -> both Postgres Wire and InfluxDB Line Protocol (ILP)

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+## RBAC in QuestBD Enterprise
+
+RBAC is the concept and an Access Control List (ACL) is how the concept is put into practice. We start with the assumption that you have a local QuestDB Enterprise binary running. If you can already access your QuestDB Enterprise deployment via QuestDB Cloud, [skip ahead to the next section](#simple-sql-setup).
+
+To activate the ACL, set two configuration options within QuestDB's `server.conf`:
+
+```java
+# Enable base authentication
+auth.enabled=true
```
glasstiger, line 42: auth.enabled=true can be omitted.
if you enable ACL, that includes auth already.

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+```sql
+CREATE USER alex_analyst WITH PASSWORD s3cr3t$$$$444!#;
+```
+
+Before we provide them their credentials, we should think more about how to provide permissions.
+
+### Group well, secure well
+
+At the moment, Alex's user has full permissions. An ACL is most helpful when permissions support granularity. With granularity, we choose what someone can and cannot access to a fine and careful degree. Right now, Alex is ultimately a second `root` user. We need to pare it down.
```
glasstiger, line 130: ???
At this point Alex should not be able to access anything.
No permissions granted yet.

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+```sql
+SELECT * FROM sensors;
+x Access denied [SELECT on sensors(ID)]
+```
+
+Not allowed!
+
+### Practicing "no"
+
+If we were to adjust the group and remove access to a column, the change takes effect the moment the permission is altered. For example, if we decide analysts do not need to see the `sensorId` column, OK - we can adjust access using `GRANT`.
```
glasstiger, line 188: GRANT -> REVOKE

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+Not allowed!
+
+### Practicing "no"
+
+If we were to adjust the group and remove access to a column, the change takes effect the moment the permission is altered. For example, if we decide analysts do not need to see the `sensorId` column, OK - we can adjust access using `GRANT`. 
+
+This is true across all contexts. If a user were to try to `JOIN` into an area where they lack permissions, then a denial will be received. If they cannot see the column, table or database, then they cannot access it via any aggregate.
+
+While the group concept is valuable for organization, direct changes to a specific user within a group may be needed. In that case, no problem. It is possible to `GRANT` access to specific users, and that access will supersede the group permission.
```
glasstiger, line 192: that access will supersede the group permission -> that access will form a union with the group permission.

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+ON readings.sensorId = sensId;
+```
+
+As above however, what if analysts lose access to `sensId`?
+
+```sql
+REVOKE SELECT ON readings(sensorId) FROM analysts;
+```
+
+On Alex's next `SELECT FROM * readings;` attempt, they will receive an error:
```
glasstiger, line 219: SELECT FROM * readings -> SELECT * FROM readings

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+```sql
+Access denied [SELECT on readings(sensorId)]
+```
+
+On Alex's broad `JOIN` query? The same thing, rejected.
+
+Alex has explicit full access to `sensors` and limited group access to `readings`. This demonstrates inherited permissions. New users without any specific `GRANT` permissions or group assignments have implicit full access. Once placed within a group or granted an explicit individual permission, the scope is then limited.
```
glasstiger, line 227: New users without any specific `GRANT` permissions or group assignments have implicit full access. -> If this is true, it is a bug which we need to fix ASAP.

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+## Implications of ILP
+
+When using Influx Line Protocol (ILP) for ingest, if `table_name` is set and does not yet exist in QuestDB, the table is created. What if, in the future, we only want new tables created over ILP?
```
glasstiger, line 231: Influx Line Protocol -> InfluxDB Line Protocol

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+GRANT CREATE TABLE OVER ILP to streamers;
+```
+
+A person who creates a table gets ALL table level permissions.
+
+A person who creates a column will get ALL column level permissions.
+
+When sending in a data stream via ILP, a person will receive both levels.
+
+This is very helpful for our `streamers`, who need to setup and tune new streams and will therefore inherit permissions to do whatever is needed within their creations. If an analyst were to want to view these new tables? They cannot, they now require specific `INSERT` access to do so.
```
glasstiger, line 255: INSERT -> SELECT

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+On Alex's broad `JOIN` query? The same thing, rejected.
+
+Alex has explicit full access to `sensors` and limited group access to `readings`. This demonstrates inherited permissions. New users without any specific `GRANT` permissions or group assignments have implicit full access. Once placed within a group or granted an explicit individual permission, the scope is then limited.
+
+## Implications of ILP
+
+When using Influx Line Protocol (ILP) for ingest, if `table_name` is set and does not yet exist in QuestDB, the table is created. What if, in the future, we only want new tables created over ILP?
+
+Consider we have a specialized data entry role we call `streamers`. These individuals setup new data streams via ILP. There is a specific permission to allow this, which when set will not allow table creates via other means:
```
glasstiger, line 233: There is a specific permission to allow this, which when set will not allow table creates via other means:
this is a bit confusing to me.
`CREATE TABLE OVER ILP` lets `streamers` to create tables via ILP, it does not restrict the user/service account to be able to create tables on the usual way.
i. e. if `streamers` has `CREATE TABLE` permission, it can also create the table in the webconsole.

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+A person who creates a column will get ALL column level permissions.
+
+When sending in a data stream via ILP, a person will receive both levels.
+
+This is very helpful for our `streamers`, who need to setup and tune new streams and will therefore inherit permissions to do whatever is needed within their creations. If an analyst were to want to view these new tables? They cannot, they now require specific `INSERT` access to do so.
+
+### Authentication, password or JWK
+
+Our walkthrough so far demonstrates basic password authentication. This is appropriate because our SQL-syntax is based on the PGWire protocol. Above, we created `sam_streamer` but did not provide Sam a password. That is because we expect Sam to authenticate via the QuestDB TCP authentication layer over-top ILP.
```
glasstiger, line 259: That is because we expect Sam to authenticate via the QuestDB TCP authentication layer over-top ILP.
can we rephrase this somehow?
PGWire also based on TCP.
Something like:
That is because we expect Sam to use keypair authentication.
QuestDB's  ILP connection does not support password authentication as the protocol itself does not provide the option.

blog/2023-08-17-enterprise-rbac-acl.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+To do so, we can `ALTER` the user and create a JWK access token:
+
+```sql
+ALTER USER sam_streamer CREATE TOKEN TYPE JWK;
+```
+
+We return the key values necessary to make an `auth.txt` file that we then reference in our QuestDB `server.conf` file. With their keys in hand, `sam_streamer` creates an authenticated stream which can create new tables in the database.
```
glasstiger, line 267: no need for the `auth.txt` file and no need for config changes.
the client can just use the JWK to connect straight away and without server restart.

goodroot:
Thanks @glasstiger ! Considered adding a blurb about service accounts. But IMO keeping this basic, then link to your article when it is complete, is a solid path forward. 

Would appreciate your thoughts on that, and a final review.

glasstiger (dismissed):

glasstiger:
> Thanks @glasstiger ! Considered adding a blurb about service accounts. But IMO keeping this basic, then link to your article when it is complete, is a solid path forward.
> 
> Would appreciate your thoughts on that, and a final review.

Added small changes and approved the PR.
If you make further edits which would dismiss my approval, just ping me and will re-approve.
Thank you.

goodroot:
Thank you @glasstiger 🎉

glasstiger (approved):

PR #1868, opened 2023-08-16
chore(docs): Change tab order in the Getting Starter - Download Binary page

jerrinot:
This is the page: https://questdb.io/docs/get-started/binaries/
Linux and Windows are the most important users of binary artifacts.

There is still an issue tho: It lists prerequisites:
image

This is valid only for the "Any (no-JVM)" version. Other versions have a JVM bundled. I suspect this might scare people off. I think the best would be to display the _Prerequisites_ paragraph(s) only when then "Any (no-JVM)" tab is selected. I tried to fix it myself, but sadly, that's beyond my React-fu.  As we use a custom component for displaying the tabs with artifacts and I could not figure out how to change it to accept children(?) with content, which should be displayed only on a specific tab.


jerrinot:
I created a new issue for the problem described above: https://github.com/questdb/questdb.io/issues/1869

PR #1873, opened 2023-08-16
chore(docs) update storage model doc

goodroot:
Clarifies `cairo.commit.mode=sync`.

Source = [Slack community](https://questdb.slack.com/archives/C1NFJEER0/p1692135616284589)


docs/concept/storage-model.md
```diff
@@ -83,6 +83,20 @@ cairo.commit.mode=sync
   width={745}
 />
 
+It is important to note that, as a result of the increase of `msync()` and `fsync()` calls, CPU usage may increase as more dirty pages are stored in memory prior to disk write. However, it is possible to tune OS-level settings to reduce the number of disk flushes that result from this increase. Doing so may help mitigate the intensity of the CPU usage increase.
```
ideoma, line 86: Sorry, this doesn't make sense to me.
With sync enabled, fewer pages will be stored in memory. I don't think there is a way to mitigate it. Reducing flushes is not an option because the flushes are **forced** by the database.

I'm not sure that the below is true either. Where did you get this info?

I think here we should just say something like 

```
It is important to note that, as a result of the increase of `msync()` and `fsync()` calls, CPU usage will increase especially with the frequent data writing commit pattern. 
```

goodroot: Yikes! Misunderstood.

Came from here: https://superuser.com/questions/1057007/adjust-linux-disk-flush-intervals-to-avoid-blocking-user-processes & https://lonesysadmin.net/2013/12/22/better-linux-disk-caching-performance-vm-dirty_ratio/ -- I'd assumed that OS level settings would ultimately govern the rate and context of when flushing occurs.

ideoma (approved):

PR #1876, opened 2023-08-17
Set max-width for inline code in tables

argshook:
Set max-width on inline code in tables to avoid very wide table columns.

Commit log:

* set max-width on inline code in tables
    this way we avoid a possible very wide table column


src/css/_global.css
```diff
@@ -273,6 +273,11 @@ html[data-theme="light"] .markdown img[src$=".svg"][alt*="chart"] {
   filter: invert(1) hue-rotate(200deg);
 }
 
+.markdown table code {
+  display: inline-block;
+  max-width: 100px;
```
oliver-daniel, line 278:
```suggestion
  max-width: 25ch;
```
This makes things slightly wider while still allowing for overflow behaviour to be minimized.

oliver-daniel (dismissed):
Fix approved with a minor suggestion.

oliver-daniel (approved):
Sorry, didn't realize that suggestions invalidated approval!

PR #1877, opened 2023-08-17
chore(docs) IPv4 clarity

goodroot:
Alters language regarding string/int & its relevance to how IPv4 is passed.


docs/reference/function/ipv4.md
```diff
@@ -6,7 +6,7 @@ description: A list of SQL functions supported by the IPv4 data type.
 
 The IPv4 datatype supports many functions.
 
-For IPv4 operators, see the [IPv4 Operators page](/docs/reference/operators/ipv4/).
+IPv4 addresses are presented between the range `0.0.0.1 - 255.255.255.255` and are stored as `int`. However, some IPv4 Operators allow IPv4 addresses to pass as `string` and contain a subnet and/or netmask. For the full list of IPv4 Operators and more detailed syntax, see the [IPv4 Operators page](/docs/reference/operators/ipv4/).
```
allegraharris, line 9:
```suggestion
For IPv4 operators, see the [IPv4 Operators page](/docs/reference/operators/ipv4/).
```

allegraharris: There aren't any general SQL functions that take a string IPv4 arg so perhaps best to revert to what it was before

docs/reference/operators/ipv4.md
```diff
@@ -6,13 +6,11 @@ description: Operator examples with IPv4 type data.
...
-IPv4 addresses are **passed** as `string`, but are **stored** as `int`.
+IPv4 addresses are presented between the range `0.0.0.1 - 255.255.255.255`.
 
-When applied as IPv4, they are presented between the range `0.0.0.1 - 255.255.255.255`.
+When passed, they may contain trailing and leading dots: `...1.1.1.1...`.
 
-When passed as a `string`, they may contain a subnet mask: `35.24.65.2/16`
-
-Addresses when passed may contain trailing and leading dots: `...1.1.1.1...`.
+IPv4 addresses are stored as `int`, but can be passed as `string`. When passed as `string`, they may contain a subnet mask (`35.24.65.2/16`) or a subnet with a netmask(`2.232/16`). See below to see which operators support `string` type arguments and which do not.
```
allegraharris, line 13:
```suggestion
IPv4 addresses are stored as `int` but are passed as `string` in dotted decimal notation.
```

allegraharris: I think it might be better to only mention netmask stuff with the specific operators/functions that support it so as to avoid confusion

allegraharris: I can't add comments on those lines as they were not changed in this pr but would it be possible to add examples of string arguments under <<. <<=, rnd_ipv4(string, int) and netmask() as those are the only operators/functions that support string args? 

Also, just noticed that we don't have netmask() on this page - would it be possible to add it please? See this pr for detail https://github.com/questdb/questdb/pull/3624

allegraharris: And also maybe an example for rnd_ipv4() as it's the only one that doesn't have one? Thanks!

allegraharris (changes requested):

docs/reference/function/ipv4.md
goodroot: Some of the function examples contain operators.

That made me feel they need some additional context.

Thoughts?

docs/reference/operators/ipv4.md
goodroot: If we don't have anything about subnets/netmasks/strings at the top of this page, we don't discuss them until they appear in the examples. I'd say it's okay to intro them, especially because they do appear in the function examples too:

```
SELECT ip, bytes, CASE WHEN ip << '2.65.32.1/2' THEN 1 ELSE 0 END FROM example
```

I'll defer to you if you feel strongly enough, but IMO this is fair balance.

---

Will also add the additions. 👍

docs/reference/function/ipv4.md
allegraharris: Yeah I see what you mean - maybe add a clarifying sentence about the fact that string args are distinct from ipv4 args that are passed as strings?

docs/reference/operators/ipv4.md
goodroot: @allegraharris Okay -- that's probably better.

WDYT?

docs/reference/function/ipv4.md
goodroot: Ah after rolling through the 3 pages, I think the advice was solid. I'll resolve this one.

docs/reference/operators/ipv4.md
```diff
@@ -219,6 +223,24 @@ Use case: calculating the range of unique addresses between two ip addresses
 ```sql
 ipv4 '92.11.8.40' - ipv4 '92.11.8.0' -> 40
 ```
+
+### Return netmask - netmask()
```
allegraharris, line 227:
```suggestion
### Return netmask - netmask(string)
```

docs/reference/operators/ipv4.md
allegraharris: Great! Thank you

allegraharris (approved):

docs/reference/sql/datatypes.md
```diff
@@ -111,26 +111,13 @@ send `UUIDs` as `strings` to be converted to UUIDs by the server.
...
-`IPv4` arguments may take the following form: 
-
-* ip address without a netmask: `23.62.200.1`
-
-`String` arguments may take the following forms:
-
-* ip address without a netmask: `23.62.200.1`
-* ip address with a netmask: `5.43.11.8/2` 
-* a subnet with a netmask: `2.232/16` 
+Within a column, all addresses are stored as a 32-bit `int`, for example: `255.255.255.255`.
```
jerrinot, line 120: > Within a column, all addresses are stored as a 32-bit int, for example: 255.255.255.255.
I don't get the example. How is `255.255.255.255` related to 32 bits int? It feels unfinished. What is the example supposed to show?

docs/reference/operators/ipv4.md
```diff
@@ -6,15 +6,20 @@ description: Operator examples with IPv4 type data.
 
 This document outlines the IPv4 data type operators.
 
-IPv4 addresses are **passed** as `string`, but are **stored** as `int`.
+IPv4 addresses are presented between the range `0.0.0.1 - 255.255.255.255`.
```
jerrinot, line 9: I would be explicit about the address `0.0.0.0` being interpreted as `null`. 
_What a strange range! Why not 0.0.0.0?_ <-- This was the first thing on my mind whilst reading this.

docs/reference/sql/datatypes.md
allegraharris: I'm unsure whether we should specify that it's stored as a 32-bit int because it might just confuse a user because no one thinks of an ip address in terms of an int. E.g., `255.255.255.255` is stored internally as -1. Do you think we should remove the example of the 32-bit int info? @jerrinot

goodroot: The maximum statement of an IPv4 address. Between `0.0.0.1-255.255.255.255`.

Would this be better?

```
They are - as one would imagine - very useful when dealing with IP addresses.

Within a column, all addresses are stored as a 32-bit `int`.

IPv4 addresses exist within the range of `0.0.0.1` - `255.255.255.255`.
```

docs/reference/operators/ipv4.md
```diff
@@ -6,15 +6,20 @@ description: Operator examples with IPv4 type data.
...
 
-When applied as IPv4, they are presented between the range `0.0.0.1 - 255.255.255.255`.
+When passed, they may contain trailing and leading dots: `...1.1.1.1...`.
```
jerrinot, line 11: do we really want to document this? what is it good for? maybe it could stay as an _undocumented implementation detail_ so we have the freedom to change it if this is not desirable for any reason?

docs/reference/operators/ipv4.md
goodroot: Good call! Adding.

docs/reference/operators/ipv4.md
goodroot: I can't think of a practical use case other.

If we were voted I'd vote "not document".

allegraharris: This was implemented because browsers are tolerant of leading and trailing dots. I'm happy to keep/remove from docs

docs/reference/sql/datatypes.md
goodroot: Or rather:

```
They are - as one would imagine - very useful when dealing with IP addresses.

IPv4 addresses exist within the range of `0.0.0.1` - `255.255.255.255`.

Within a column, all addresses are stored as a `int`.
```

If we'd rather never reference the bits, makes sense to me.

docs/reference/operators/ipv4.md
goodroot: Huh - TIL! 

Okay, pushed up.

docs/reference/operators/ipv4.md
```diff
@@ -102,7 +107,7 @@ Use case: testing to see if one IP address is not equal to another.
 ipv4 '44.8.9.10' != ipv4 '6.2.90.1' -> T
```
jerrinot, line 107: not directly related to this PR, but perhaps worth fixing too. 
this section is non-obvious. for 2 reasons:
## the first reason
`ipv4 '44.8.9.10' != ipv4 '6.2.90.1' -> T`
What is `T`? I figured out it's `true`, but it took me some effort to realize it. That's not good. 

Compare IPv4 example:
image


With an existing example at https://deploy-preview-1877--unruffled-blackwell-31bfb2.netlify.app/docs/reference/operators/comparison/ 
image

Differences: The 2nd example does not force me to think. It's very obvious. Also, it gives me a proper SQL I can actually execute. Instead of an abstract syntax. I think that makes it better documentation. 

## The 2nd reason
How is IP address comparison defined? I am too ignorant to be aware of any IP comparison standard. And my intuition is failing me: `127.0.0.1 > 127.0.1.0` is this `true` or `false`? I am not sure.

goodroot: 👍 Saved these notes for new PR, as broad pass is needed to address many quality-of-life changes.

Thanks!

docs/reference/sql/datatypes.md
jerrinot: I would simplify it furher:
```
They are useful when dealing with IP addresses. 
The  IP addresses can be in the range of `0.0.0.1` - `255.255.255.255`. The address - `0.0.0.0` is interpreted as null.
```

goodroot: 💥

jerrinot (dismissed):
good improvements, thanks to both @allegraharris and @goodroot !

docs/reference/operators/ipv4.md
```diff
@@ -219,6 +223,24 @@ Use case: calculating the range of unique addresses between two ip addresses
 ```sql
 ipv4 '92.11.8.40' - ipv4 '92.11.8.0' -> 40
 ```
+
+### Return netmask - netmask(string)
+
+Takes a `string` argument in the format `22.59.138.9/8`.
```
allegraharris, line 229: I think we should note that we also accept subnet + netmask e.g., `2.2/16'

allegraharris (dismissed):

goodroot:
@jerrinot Hey! Mind re-approving?

bluestreak01 (approved):

PR #1880, opened 2023-08-18
fix(): remove double "protocol" word

argshook:
Fixed the issue with automated ILP expansion duplicating the word 'protocol'.

Commit log:

* fix automated ILP expansion to not say 'protocol' twice


puzpuzpuz (approved):
Good spot :)

PR #1887, opened 2023-08-22
chore(blog) Fix plausible prop

goodroot:
Only one = is required.


oliver-daniel (approved):
Searching for all these variants of QuestDB makes me think of: https://www.youtube.com/watch?v=31euOK67UrM

PR #1888, opened 2023-08-22
chore(all) Return of Posthog

goodroot:
After researching and getting a better understanding of Docusaurus, I cannot determine a reason why Posthog would have caused SEO issues.

In addition, after experimenting with Plausible event and page tracking, my feeling is that Posthog is an advanced solution we'd be very lucky to have. Especially as it bridges .com & .io to great effect.

Happy to provide a more nuanced argument, but this PR restores based on prior removal & clean PRs:

1. https://github.com/questdb/questdb.io/pull/1698
2. https://github.com/questdb/questdb.io/pull/1783

Prior to merge, requires env vars to be reset within Netlify.


goodroot:
NO! 💀

PR #1889, opened 2023-08-23
chore(all) Analytics pass

goodroot:
* applies advice from semrush: broken links, descriptions for glossaries, search page
* plausible tracking to all outbound cloud ctas: cloud top/bottom, main fold, primary & pricing
* enables deeper outbound tracking & pageview capabilities (not implemented) from plausible
* extends the cloudURL component to optionally take UTM information, tests on blogCTA


goodroot:
Screenshot 2023-08-23 at 1 24 03 PM

Not that after a successful test, I have removed the `blog-cta` source UTM tag. Will apply them across the board once a strategy can be considered. For now, consider this just the wiring to make it possible. This will effectively bridge this site with posthog.

src/modules/cloud/Top/index.tsx
```diff
@@ -61,6 +61,7 @@ export const Top = () => {
           className={styles.startBuildingButton}
           to={cloudUrl}
           newTab={false}
+          plausible-event-name="Click+Button+Cloud+Top"
```
argshook, line 64: i have a few questions about this attribute:

1. according to [this doc](https://plausible.io/docs/custom-props/for-custom-events) the custom events should be passed as a CSS class. Will it really work as a custom attribute? Maybe you have another source with more info?
2. the attribute `plausible-event-name` is not implemented in `Button` component, so it doesn't get rendered in the DOM.
2. ideally, if custom attribute is really supported by plausible (question `1.`) then maybe we are able to tell plausible which attribute to look at? If yes, we should use `dataHook` attribute: it is something we already use in tests to attach to a specific DOM node and it's already added to `Button` component.

goodroot: Happy to expand to a better implementation.

A test event went in an earlier PR, and it is indeed functional:

Screenshot 2023-08-24 at 8 47 22 AM

&&

Screenshot 2023-08-24 at 8 48 30 AM

goodroot: @argshook Ok that should do it -- learning my way around React. Thanks for patience. 

Decided not to add to button component, and do it on each Cloud call for now. In the future, I can see a need to add "buckets" of events, for Cloud, or Enterprise, Demo, Contact, etc. That be a fair time to make the button smarter.

src/modules/pricing/get-started/index.tsx
```diff
@@ -7,7 +8,11 @@ import styles from "./styles.module.css"
 
 export const GetStarted = () => (
   <Section center fullWidth noGap className={styles.root}>
-    <Button size="small" to={customFields.cloudUrl}>
+    <Button
+      className={clsx("plausible-event-name=Click+Button+Pricing")}
```
argshook, line 12:
```suggestion
      className="plausible-event-name=Click+Button+Pricing"
```

src/pages/cloud/index.tsx
```diff
@@ -73,7 +74,12 @@ const CloudPage = () => {
           <Section.Title center>Ready to get started?</Section.Title>
 
           <Section noGap center>
-            <Button variant="primary" to={cloudUrl} newTab={false}>
+            <Button
+              className={clsx("plausible-event-name=Click+Button+Cloud+Bottom")}
```
argshook, line 78:
```suggestion
              className="plausible-event-name=Click+Button+Cloud+Bottom"
```

src/pages/glossary/change-data-capture.md
```diff
@@ -85,31 +86,34 @@ to track and capture changes:
   the logs. However, decoding the transaction log may be complex and requires
   additional system components outside the database.
 
-    <Screenshot
+  {" "}
```
argshook, line 89:
```suggestion
```

argshook (dismissed):

argshook (approved):

PR #1896, opened 2023-08-29
feat(*): New `Get QuestDB` page

insmac:


goodroot:
Nice one! This page sure fills a gap...

Did not alter matrix until we're aligned on format in Slab.

A couple minor visual things:

---

Contrast on green version text on light mode is bright.

Screenshot 2023-08-29 at 10 15 35 AM

Also, on light, the 3-pack feels very "in the open".

Would we consider a container, like with these?

Screenshot 2023-08-29 at 10 21 16 AM

---
I'm trying to find a way to point out "no CC trial". Its not something I see going in the descriptions, and I don't want to cram the _$200 free credit_ pill.

Any thoughts?

Screenshot 2023-08-29 at 10 32 55 AM

---

There are two reasons that I think the `/open-source/` page may be better as `/downloads`: 

* can see "open source" getting its own landing page at some point
* SEO -- "download questdb" is a popular query that we expect will increase -- hitting that "on the nose" with a download page would serve us well.

---

Also much appreciate the quick work getting this page in code. 🔥

insmac:
@goodroot Addressed the issues with the light theme, and swapped the URL to `/download/`. 

As for "no card required", I propose the following approach:
CleanShot 2023-08-30 at 12 23 08@2x

To maintain symmetry, I have moved the blue links below the description, but above the main CTA, so that all content rows are equally distributed.

Let me know what you think 🙂 (the above is a screenshot from Figma)

goodroot:
@insmac Nice one!

I'd say it's worth putting the product version in navbar + the altered button in code.

